PR #8, opened 2023-12-22
LNK-1584: BOTW Submission Bundle - Individual measure report structure is very different from MVP

SaraMunich-Lantana:
added some code to remove the extensions that reference the resources that had been removed from contained


Report/Core/MeasureReportSubmissionBundler.cs
```diff
@@ -200,6 +200,7 @@ private void BundleSharedPatientLineLevel(Bundle bundle, MeasureReport measureRe
                 measureReport.Contained.RemoveAt(0);
 
                 // remove the ext that reference the contained resource
+                measureReport.Extension.RemoveAll(e => e.Value is Extension);
```
arianamihailescu, line 203: We should not remove all from Extension, only the ones that reference the contained resources.

arianamihailescu (approved):

PR #9, opened 2024-01-04
Period dt fix operation splitout

edward-miller-lcg:


Normalization/Application/Commands/PeriodDateFixerCommand.cs
```diff
@@ -0,0 +1,49 @@
...
+    public Bundle Bundle { get; set; }
+    public List<PropertyChangeModel> PropertyChanges { get; set; }
+}
+
+public class PeriodDateFixerCommandHandler : IRequestHandler<PeriodDateFixerCommand, OperationCommandResult>
+{
+    public async Task<OperationCommandResult> Handle(PeriodDateFixerCommand request, CancellationToken cancellationToken)
+    {
+        var bundle = request.Bundle;
+        var propertyChanges = request.PropertyChanges;
```
amphillipsLGC, line 20: Doesn't seem like this is needed or maybe it will get some use later?

amphillipsLGC (approved):
Looks good!

Normalization/Application/Commands/PeriodDateFixerCommand.cs
edward-miller-lcg: fixed.

amphillipsLGC (approved):

PR #14, opened 2024-01-12
LNK-1584:  Individual measure report structure is very different from…

arianamihailescu:
… MVP


dvargaslantana (approved):
looks good! Not necessary for this PR, but wondering if at some point we add these extensions in a db table rather than in the code.

PR #15, opened 2024-01-19
LNK-1876: Add ability to connect to Event Hub in the Audit Service

amphillipsLGC:
Added temporary kafka connection model to allow testing against event hubs. If all works well this can be refactored into the shared service.


dvargaslantana (approved):
looks good!

PR #17, opened 2024-02-07
LNK-1913: Update audit and notification services to dot net8

amphillipsLGC:
- Updated audit service to .NET 8
- Updated notification service to .NET8
- Update telemetry
- overhauled logging in both services to use `MessageLogger` and log properties attributes
- Cleaned up logging to reduce chattiness 
- Added logging redactors (star redactor, and hmac redactor) using data taxonomy attributes
- Overhauled configuration to be able to leverage auzre app config and other external configuration services


Audit/Presentation/Services/AuditService.cs
dvargaslantana, line 1: Does this file need to exist?

Audit/Program.cs
```diff
@@ -126,11 +157,29 @@ static void RegisterServices(WebApplicationBuilder builder)
...
+    builder.Services.AddRedaction(x => {
+
+        x.SetRedactor<StarRedactor>(new DataClassificationSet(DataTaxonomy.SensitiveData));
+
+        var hmacKey = builder.Configuration.GetValue<string>("Logging:HmacKey");        
+        if (!string.IsNullOrEmpty(hmacKey))
+        {           
+            x.SetHmacRedactor(opts => {
+                opts.Key = Convert.ToBase64String(Encoding.UTF8.GetBytes(hmacKey));
+                opts.KeyId = 808;
```
dvargaslantana, line 171: Maybe not worth it now or what the standard practice is for this, but may want to enumerate or keep a dictionary of the id's if we decide to use this cross service

dvargaslantana (approved):
looks good!

Audit/Program.cs
amphillipsLGC: Yeah I agree we should standardize it. Maybe we can look into if there is any best practice for the Ids, but I believe it is just a design decision for our team.

Audit/Presentation/Services/AuditService.cs
amphillipsLGC: No good catch, I forgot to remove it after testing that it worked while commented out. I removed all references to the create audit endpoints since the service should only create audit logs from consumed events.

dvargaslantana (approved):

PR #19, opened 2024-02-13
LNK-1869: Add entity framework to audit service

amphillipsLGC:
- Added entity framework 8
- Created new `DbContext` and `AuditLogRepository`
- Refactored Audit entity model
- Added strongly typed Audit Id
- Refactored endpoints, command, and queries
- Added scoped service factory to deal with issues around the `AuditEventListener` being a singleton background service and `DbContext` being a scoped service.
- Added new `DatabaseProvider` configuration property


Audit/Persistance/Repositories/AuditLogRepository.cs
```diff
@@ -0,0 +1,120 @@
...
+            }
+
+            if (filterUserBy is not null && filterUserBy.Length > 0)
+            {
+                query = query.Where(x => x.User == filterUserBy);
+            }
+
+            if (sortBy is not null && sortBy.Length > 0)
+            {
+                query = sortBy switch
```
dvargaslantana, line 87: Maybe apply a toLower() to the comparison if the search filters are publicly accessible?

dvargaslantana (approved):
looks good

Audit/Persistance/Repositories/AuditLogRepository.cs
amphillipsLGC: I think this is a good thing to look into. SQL Server is case insensitive by default, but if we switched providers to Postgres it would be case sensitive. Will create a ticket to look into the performance impact of this.

PR #28, opened 2024-02-23
Update all Projects to .NET 8 and Update NuGets as Needed

MontaltoNick:
Update Normalization and QueryDispatch to .NET 8


amphillipsLGC (approved):
Looks good!

amphillipsLGC (approved):
We'll have to make sure to coordinate with DevOps for the necessary changes around https://learn.microsoft.com/en-us/dotnet/core/compatibility/containers/8.0/aspnet-port

PR #29, opened 2024-02-23
LNK-2090: Updating audit service configuration settings

amphillipsLGC:
- Updated configuration properties to better support database connection strings in azure app config
- Fixed unit test errors
- Added ability to set sort order in audit searches
- Added the Resource property to free text search
- Refactored the audit metrics to use an interface 
- Fixed an issue where the action was not getting saved correctly when consuming an auditable event
- code cleanup


dvargaslantana (approved):
looks good!

PR #62, opened 2024-03-20
LNK-2206-Audit-log-database-performance-adjustments

amphillipsLGC:
- Added an identity column to act as the primary key for the audit log table
- added additional indexes
- refactored the search method into it's own repository to allow for optimized provider specific search method
- fixed unit tests
- Updated NuGet packages
- Added entity framework open telemetry instrumentation


arianamihailescu (approved):
Looks good to me

PR #63, opened 2024-03-20
LNK-2106

nvmLantana:
Move Exception Handling into Shared


Shared/Application/Error/Handlers/TransientExceptionHandler.cs
```diff
@@ -0,0 +1,91 @@
...
+        protected TransientExceptionHandler(ILogger<TransientExceptionHandler<K, V>> logger,
+            IKafkaProducerFactory<string, AuditEventMessage> auditProducerFactory,
+            IKafkaProducerFactory<K, V> producerFactory)
+        {
+            Logger = logger;
+            AuditProducerFactory = auditProducerFactory;
+            ProducerFactory = producerFactory;
+        }
+
+        public virtual void HandleException(ConsumeResult<K, V> consumeResult, Exception ex)
```
edward-miller-lcg, line 36: Would it make sense to pass the topic name as a method argument here and pass it down as opposed to setting it in the calling class?

edward-miller-lcg (commented):
@dvargaslantana
I don't see NonTransientException handling implemented for deadletter workflow so that would need to be added either before the PR is approved or as an after task.

MontaltoNick:
> @dvargaslantana I don't see NonTransientException handling implemented for deadletter workflow so that would need to be added either before the PR is approved or as an after task.

ExceptionHandler is the workflow for the 'non transient' exception case. It just logs the audit message, and back in the service we mark the event as processed so it is removed from the queue.

edward-miller-lcg:
I think we should be putting that in a specific non transient case that produces a dead letter. I don't understand why we would not do that for NonTransient but _do_ that for Transient cases.

edward-miller-lcg (approved):

PR #79, opened 2024-04-01
LNK-2159: Shared -Add- SQL - Persistence

arianamihailescu:


Shared/Shared.csproj
```diff
@@ -14,7 +14,7 @@
 		<Description>Shared library for Lantana Link</Description>
 		<Product>NHSN Link Shared Library</Product>
 		<!--<Version>1.1.1</Version>-->
-		<VersionPrefix>1.2.1</VersionPrefix>
+		<VersionPrefix>1.2.2</VersionPrefix>
```
edward-miller-lcg, line 17: So it looks like this would be a minor version change. So 1.3.1

edward-miller-lcg (changes requested):
See comments

edward-miller-lcg (approved):

edward-miller-lcg (approved):

PR #82, opened 2024-04-03
LNK-2240: Implement security in botw validation service and measure eval services

amphillipsLGC:
- Added dependency on `spring-boot-starter-oauth2-resource-server`
- Updated security helper to use Oauth with JWT authentication
- Added policy to allow anonymous access to the health check endpoint
- Added api to the base route to match other services


smailliwcs (approved):
Approving, but something to consider ...

Do we want to deactivate this by default for local development?  Or provide a Compose file that spins up a properly configured Keycloak instance?  (Or maybe an internally available config file that points to `oauth.nhsnlink.org`?)  Because, as currently implemented, you can't just pull `measureeval` or `validation` and run it ... Spring will try (and fail) to hit an OAuth server at `localhost:7080`.  IMO, this is something we need to address more generally ... reducing the distance between a clean checkout and a running service (locally or otherwise).  But not necessarily something we address as part of this PR.

amphillipsLGC:
> Approving, but something to consider ...
> 
> Do we want to deactivate this by default for local development? Or provide a Compose file that spins up a properly configured Keycloak instance? (Or maybe an internally available config file that points to `oauth.nhsnlink.org`?) Because, as currently implemented, you can't just pull `measureeval` or `validation` and run it ... Spring will try (and fail) to hit an OAuth server at `localhost:7080`. IMO, this is something we need to address more generally ... reducing the distance between a clean checkout and a running service (locally or otherwise). But not necessarily something we address as part of this PR.

That is a good point. Let me see if there is something something similar in spring boot that allows you to not use it if you are in a development environment.

PR #87, opened 2024-04-04
Supporting runtime config in Demo App

seanmcilvenna:
Adding proxy to angular to allow angular to proxy /proxy-api/ to localhost:7777 to avoid cors when testing

Moving `environment` to `AppConfigService`
`AppConfigService` loads settings from a runtime `/assets/app.config.json` and optionally `/assets/app.config.local.json`


amphillipsLGC (approved):
looks good

PR #89, opened 2024-04-05
LNK-2159: Shared refactor

arianamihailescu:


Shared/Domain/Entities/BaseEntity.cs
dvargaslantana, line 1: I suspect this will cause conflicts with services that will still be using MongoDB as their repo (Report service for example)

arianamihailescu: No, cause is still string and the value is supplied by the code, we do not use annotations to generate ObjectId but we set the id with a string representation of a Guid. I also tested with Report and Measure Eval that both use MongoDb.

dvargaslantana (approved):
looks good

PR #90, opened 2024-04-08
LNK-2193

dvargaslantana:
- Added retry listener
- Added dead letter and retry workflow to patientEventListener

### DIFF
--- a/QueryDispatch/Listeners/PatientEventListener.cs
+++ b/QueryDispatch/Listeners/PatientEventListener.cs
@@ -10,6 +10,8 @@ using LantanaGroup.Link.Shared.Application.Interfaces;
 using System.Text;
 using LantanaGroup.Link.Shared.Application.Models.Kafka;
 using QueryDispatch.Application.Settings;
+using LantanaGroup.Link.Shared.Application.Error.Interfaces;
+using LantanaGroup.Link.Shared.Application.Error.Exceptions;
 
 namespace LantanaGroup.Link.QueryDispatch.Listeners
 {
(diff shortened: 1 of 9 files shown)


QueryDispatch/Listeners/PatientEventListener.cs
```diff
@@ -61,29 +78,26 @@ protected override Task ExecuteAsync(CancellationToken stoppingToken)
...
+                                    Value = Encoding.UTF8.GetString(e.ConsumerRecord.Message.Value),
+                                    Headers = e.ConsumerRecord.Message.Headers
+                                }
                             };
 
-                            ProduceAuditEvent(auditValue, e.ConsumerRecord.Message.Headers);
+                            _consumeResultDeadLetterExceptionHandler.HandleException(converted_record, new DeadLetterException("Consume Result exception: " + e.InnerException.Message, AuditEventType.Create), facilityId);
 
-                            //TODO: We will eventually have Error/Retry workflows implemented
                             _patientEventConsumer.Commit();
```
smailliwcs, line 94: Should we be calling `Commit(e.ConsumerRecord)` here?  (Does that even work?)  Could this arg-less `Commit` potentially commit more offsets than we want?

amphillipsLGC (approved):
Looks good

QueryDispatch/Listeners/PatientEventListener.cs
dvargaslantana: This is a good point. From what I've seen so far, you can't commit the ex.ConsumeRecord from the exception. Will dig into this a bit.

dvargaslantana: Something like this may work out. 
`_patientEventConsumer.Commit(new List<TopicPartitionOffset> { e.ConsumerRecord.TopicPartitionOffset });`
Testing it out now.

QueryDispatch/Listeners/PatientEventListener.cs
```diff
@@ -61,29 +78,26 @@ protected override Task ExecuteAsync(CancellationToken stoppingToken)
...
-                                Action = AuditEventType.Query,
-                                ServiceName = "QueryDispatch",
-                                EventDate = DateTime.UtcNow,
-                                Notes = $"Kafka PatientEvent consume failure, potentially schema related \nException Message: {e.Error}",
+                                Message = new Message<string, string>()
+                                {
+                                    Key = facilityId,
+                                    Value = Encoding.UTF8.GetString(e.ConsumerRecord.Message.Value),
+                                    Headers = e.ConsumerRecord.Message.Headers
+                                }
```
smailliwcs, line 89: Probably want some null checking around the message key and value.  A null `e.ConsumerRecord.Message.Key`/`Value` here will cause `Encoding.UTF8.GetString` to throw a `NullReferenceException` and bring down the service.

PR #91, opened 2024-04-08
Lnk-2162 Entity Framework Census Service

edward-miller-lcg:


arianamihailescu (changes requested):
Please fix the compile errors.

Census/Migrations/20240408185717_InitialCreate.cs
```diff
@@ -0,0 +1,82 @@
...
+    public partial class InitialCreate : Migration
+    {
+        /// <inheritdoc />
+        protected override void Up(MigrationBuilder migrationBuilder)
+        {
+            migrationBuilder.CreateTable(
+                name: "CensusConfig",
+                columns: table => new
+                {
+                    Id = table.Column<string>(type: "nvarchar(450)", nullable: false),
```
arianamihailescu, line 18: One thing I did different in Tenant was to overwrite the BaseEntity from Shared and replace String with Guid so that column gets generated with type uniqueIdentifier versus varchar like:
CREATE TABLE [Facilities] (
     [Id] uniqueidentifier NOT NULL,

Census/Census.csproj
```diff
@@ -66,6 +65,11 @@
 
   <ItemGroup>
     <Folder Include="Listeners\" />
+    <Folder Include="Migrations\" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Shared\Shared.csproj" />
   </ItemGroup>
```
arianamihailescu, line 73: Should we include the shared library there instead of the Shared project directly?

arianamihailescu (approved):

PR #111, opened 2024-04-16
Java shared project no longer using maven feed

seanmcilvenna:
* Shared is directly referenced as a project
* Root POM includes dependencies that are used by both measureeval and validation services
* Added repository for measure definitions to measureeval
* Added REST controller for measure definitions and tested that mongo connection works
* Dockerfiles updated to build both destination service and shared module at same time

### DIFF
--- a/Java/measureeval/Dockerfile
+++ b/Java/measureeval/Dockerfile
@@ -2,17 +2,14 @@ ARG SHARED_BOTW_FEED_TOKEN_ARG
 
 # Stage 1: Build the Spring Boot application
 FROM maven:3.8.7-openjdk-18-slim AS build
-ARG SHARED_BOTW_FEED_TOKEN_ARG
 WORKDIR /app
-COPY . /app
-ENV SHARED_BOTW_FEED_TOKEN=$SHARED_BOTW_FEED_TOKEN_ARG
-COPY .m2/settings.xml /root/.m2/settings.xml
-RUN mvn clean install -DskipTests
-RUN mv target/measureeval-*.jar target/measureeval.jar
+COPY Java/ ./
+RUN mvn clean install -pl measureeval -am
+RUN mv measureeval/target/measureeval-*.jar measureeval/target/measureeval.jar
 
 # Stage 2: Run the Spring Boot application
 FROM openjdk:17-jdk-alpine
 WORKDIR /app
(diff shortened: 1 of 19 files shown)


Java/measureeval/Dockerfile
```diff
@@ -2,17 +2,14 @@ ARG SHARED_BOTW_FEED_TOKEN_ARG
...
+RUN mv measureeval/target/measureeval-*.jar measureeval/target/measureeval.jar
 
 # Stage 2: Run the Spring Boot application
 FROM openjdk:17-jdk-alpine
 WORKDIR /app
 COPY --from=build /app/target/measureeval.jar measureeval.jar
 EXPOSE 8080
-CMD ["java", "-jar", "measureeval.jar"]
\ No newline at end of file
+CMD ["java", "-jar", "measureeval.jar", "--add-opens java.base/java.util=ALL-UNNAMED"]
```
smailliwcs, line 15: Just curious ... is this something Mongo needs for serialization?

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/controllers/HealthController.java
```diff
@@ -5,7 +5,7 @@
 import org.springframework.web.bind.annotation.RestController;
 
 @RestController
-@RequestMapping("/_health")
+@RequestMapping("/health")
```
smailliwcs, line 8: Do we have an project-wide convention to put this at `/health` rather than `/_health`?  `SecurityHelper` seems to assume it's at `/_health`.  Regardless, should probably put the correct value in a constant and reference that where needed.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/entities/MeasureDefinition.java
```diff
@@ -0,0 +1,30 @@
...
+    private String id;
+
+    @Field("measureDefinitionId")
+    private String measureId;
+
+    @Field("measureDefinitionName")
+    private String measureName;
+
+    @Field("url")
+    private String measureUrl;
```
smailliwcs: Any reason _not_ to have the Java field names match the Mongo property names?

smailliwcs (approved):
Dropped a few comments for further discussion, but gonna approve and merge so I can pull into my work.

PR #118, opened 2024-04-23
LNK-2362: Instrument remaining consumers and producers

amphillipsLGC:
Added telemetry instrumentation for consumers in the query dispatch, report, and submission services.


dvargaslantana (approved):
looks good

PR #137, opened 2024-04-27
LNK-2376: Update shared project telemetry (Diagnostic names, extension methods)

amphillipsLGC:
- Added `DiagnosticNames` class for consistent names used in telemetry across services
- Updated existing service metrics to use new diagnostic names class.
- Fixed bug that was causing the custom traces to not be captured. The service name int he activity source and the service name in the open telemetry configuration were not matching for most services.
- Fixed a configuration issue that was preventing link admin from accepting bearer tokens created by link admin

### DIFF
--- a/DotNet/LinkAdmin.BFF/Program.cs
+++ b/DotNet/LinkAdmin.BFF/Program.cs
@@ -42,6 +42,11 @@ app.Run();
 #region Register Services
 static void RegisterServices(WebApplicationBuilder builder)
 {
+
+    //Initialize activity source
+    var version = Assembly.GetExecutingAssembly().GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion ?? "1.0.0";
+    ServiceActivitySource.Initialize(version);
+
     // load external configuration source if specified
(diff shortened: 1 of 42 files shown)


DotNet/LinkAdmin.BFF/Program.cs
```diff
@@ -42,6 +42,11 @@
 #region Register Services
 static void RegisterServices(WebApplicationBuilder builder)
 {
+
+    //Initialize activity source
+    var version = Assembly.GetExecutingAssembly().GetCustomAttribute<AssemblyInformationalVersionAttribute>()?.InformationalVersion ?? string.Empty;
```
seanmcilvenna, line 47: Should this be applied to all services? I only see this change in `Account/Program.cs` right now...

nvmLantana (approved):
Looks good from a Report/Submission/Shared perspective

seanmcilvenna (approved):

DotNet/LinkAdmin.BFF/Program.cs
amphillipsLGC: I only implemented it in LinkAdmin for now to get feedback on the approach. I guess I probably should have removed it or just added it to all the services.

PR #138, opened 2024-04-29
LNK-2193

dvargaslantana:
Updated Program.cs to load the listeners based on the set ConsumerSetting values


amphillipsLGC (approved):
Looks good!

PR #141, opened 2024-04-30
Auto migrate extension

seanmcilvenna:
Adding app extension method to auto migrate EF for a DbContext
Applying auto migrate to the Census service, other services to come later


edward-miller-lcg (changes requested):
target branch needs to be changed to `dev`

edward-miller-lcg (approved):

PR #152, opened 2024-05-02
Updating Dockerfiles to remove use of APK

seanmcilvenna:
APK fails to run in other environments because of proxy rules, and is mostly unnecessary for these builds anyways.
Converted PatientsToQuery to use Ubuntu with APT instead, which is whitelisted in other environments already.


edward-miller-lcg (commented):
I'm worried about removing:

```
ENV \
    DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=false \
    LC_ALL=en_US.UTF-8 \
    LANG=en_US.UTF-8

```
When this isn't there, it tends to break Entity Framework. Alpine removes this that breaks EF. @amphillipsLGC

seanmcilvenna:
I can re-add that ENV statement to them... That's not the one that would break anything. I just thought it was a copy-paste for the APK.

seanmcilvenna:
@edward-miller-lcg I've re-added those ENV lines

dvargaslantana (approved):

PR #166, opened 2024-05-08
Lnk 2367 ef query dispatch

edward-miller-lcg:
- Add EF to QueryDispatch service
- Update Data Acquisition so that health check inspects sql server


nvmLantana (approved):

dvargaslantana:
Pulled down the PR and getting the following warnings.:
image

Not sure if it is something we should address. Testing the service now to make sure it's all good

dvargaslantana:
Getting an error when the QueryDispatch Job attempts to delete a PatientDispatch record 
image

PR #175, opened 2024-05-10
Improve FHIR ID deserialization

smailliwcs:
Support IDs with or without a `{type}/` prefix (as well as any other variant that HAPI's `IdType` understands).


amphillipsLGC (approved):
Looks good!

PR #179, opened 2024-05-10
DataAcquisition: Added PatientId to SearchFhirEndpointAsync for search queries

dvargaslantana:


dvargaslantana:
Not needed. Didn't update tables locally which was causing the error I was trying to fix here.

PR #180, opened 2024-05-13
LNK-2387: Update account service to support authentication and authorization

amphillipsLGC:
Reconfigured account service schema
- User → Can have many roles
- Role → Can have many users
- Role → Can have many claims
- Added User and Role Repositories
- Added commands and queries for both users and roles
- Added new user and role endpoints with request handlers
- Added fluent validation
- Added source generated logging and logging redactors
- Added ability to clear account cache when users or roles have been updated to force immediate claim updates

Updated shared service
- Added validation filter and validation exception middleware
- Added telemetry extension `StartActivityWithTags`
- Added problem details service extension
- Added sort order enum
- Added `IPagedModel` and `IPaginationMetadata` interfaces
- Added validation failure and pagination metadata response models
- Anonymous access has been added to link admin bff and to the shared service authentication extensions
- Local secret manager added
- Added ability to configure data protection

Updated link admin to use new account service schema
- Updated account service swagger endpoint to be reached through link admin
- You can authenticate into link admin using JWT tokens (oauth.nhsnlink) to then generate link tokens for backend services
- Authorization has been added to the account service
- Added ability to configure link token lifespan when generated from REST endpoint

Link.Authorization
- Added `LinkPermissions`
- Updated link authorization policies


DotNet/Audit/appsettings.json
```diff
@@ -40,7 +40,7 @@
     "OtelCollectorEndpoint": "",
     "EnableAzureMonitor": false
   },
-  "EnableSwagger": false,
+  "EnableSwagger": true,
```
seanmcilvenna: We should default "EnableSwagger" to `false` always. And require it to be overridden in lower environments. Don't want this default to accidentally lead to swagger in STAGE/PROD

DotNet/Link.Authorization/Permissions/LinkPermissions.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Link.Authorization.Permissions
+{    
+    public enum LinkPermissions
```
seanmcilvenna: We definitely need some documentation in Confluence on _this_ implementation, accompanied by a video recording. What are the implications of each of these permissions, for example? If I don't have `CanAdministerAllTenants`, I have to be assigned a tenant to see it?

seanmcilvenna (approved):

DotNet/Audit/appsettings.json
amphillipsLGC: that is a good idea

DotNet/Link.Authorization/Permissions/LinkPermissions.cs
amphillipsLGC: Correct, the idea is that if you don't have this claim, you can only administer facilities (tenants) that you have been assigned to. I do need to update a lot of documentation, but this is also a bit fluid still. I have no doubts the authorization matrix will change as we make changes to the application.

DotNet/Audit/appsettings.json
amphillipsLGC: done

PR #183, opened 2024-05-13
LNK-2041

nvmLantana:
More Robust DeadletterExceptionHandler functions for various scenarios around sending a NULL consumeResult or no ConsumeResult to the exception handler.


amphillipsLGC (approved):
looks good

PR #195, opened 2024-05-16
LNK-2445: Update link bearer services authority to be configurable

amphillipsLGC:
Made updates to allow for the configuration of the authority for the link bearer service. This change was needed to address non development environments requiring https meta data for the authority.


dvargaslantana (approved):
looks good

PR #206, opened 2024-05-20
LNK-2441 Additional Validity Checks - Normalization

edward-miller-lcg:
Bug fixing for deadletter flow. Added extra checks to trigger deadletter workflow.


arianamihailescu (approved):
Seems fine for the intent of the ticket. I understood from Ed that there needs to be a bigger refactor here on how exceptions are bubbled up and handled but that's out of scope for this bug ticket.

PR #216, opened 2024-05-22
Standardizing use/configuration of swagger and service information

seanmcilvenna:
All .NET services using the same extensions/configuration for swagger that load when "EnableSwagger" is true in config, but defaults to false.
Fixed Java services to load load swagger correctly and added swagger definitions for REST operations.
Extra: added loki/logging to measure eval service


dvargaslantana (approved):

seanmcilvenna:
Made some additional enhancements. Once approved/merged, swagger should be consistent across all services.

edward-miller-lcg (approved):

PR #217, opened 2024-05-23
LNK-2459: PatientsToQuery replacement

dvargaslantana:
This is work to remove the PatientsToQuery service from Link Cloud. Instead, the Report service will handle the responsibility of managing the list that the PatientsToQuery service was producing. I have removed the PatientsToQuery listener in this PR. I did not delete the project file in case there is CICD pipeline work needed before removing it from the solution.


seanmcilvenna (changes requested):
Suggest including the removal of the PatientsToQuery project folder, entry in the SLN, and associated /Azure_Pipelines pipelines.

dvargaslantana:
> Suggest including the removal of the PatientsToQuery project folder, entry in the SLN, and associated /Azure_Pipelines pipelines.

Okay, will include that as well. Will need close review to azure pipeline changes

dvargaslantana:
Re-requesting review @seanmcilvenna

seanmcilvenna (dismissed):

nvmLantana (approved):

PR #231, opened 2024-05-24
LNK-2459 - Removing PatientsToQuery retry reference

dvargaslantana:
Removed PatientsToQuery from Report retry listener. Removed PatientsToQuery from shared KafkaTopic enum.


arianamihailescu (approved):
Looks good

PR #232, opened 2024-05-24
Lnk 2356 enable authentication in java

arianamihailescu:


Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/auth/JwtAuthenticationEntryPoint.java
```diff
@@ -0,0 +1,20 @@
...
+import jakarta.servlet.http.HttpServletResponse;
+import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.web.AuthenticationEntryPoint;
+import org.springframework.stereotype.Component;
+
+import java.io.IOException;
+import java.io.PrintWriter;
+
+@Component
+public class JwtAuthenticationEntryPoint implements AuthenticationEntryPoint {
```
seanmcilvenna: These `com.lantanagroup.link.measureeval.auth` classes should be moved into `Shared` so that they can be re-used by both `measureeval` and `validation`.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/configs/SecurityConfig.java
```diff
@@ -1,15 +1,44 @@
...
+  @Autowired
+  private JwtAuthenticationEntryPoint point;
+
+  @Autowired
+  private JwtAuthenticationFilter authFilter;
+
+  @Bean
+  public SecurityFilterChain securityFilterChain (HttpSecurity http) throws Exception {
+
+    http.csrf(AbstractHttpConfigurer::disable).addFilterBefore(authFilter, UsernamePasswordAuthenticationFilter.class)
```
seanmcilvenna: Can this be abstracted into `shared` so that this configuration can be re-used by validation?

Java/measureeval/pom.xml
```diff
@@ -85,6 +86,32 @@
             <scope>runtime</scope>
         </dependency>
 
+        <dependency>
```
seanmcilvenna: `/Java/pom.xml` should have `<dependencyManagement>` entries for these packages that specify the version, so that all Java services use the same versions of dependencies.
Then, the `<version>` can be removed from these entries in `/Java/measureeval/pom.xml`.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/controllers/MeasureDefinitionController.java
```diff
@@ -34,17 +38,20 @@ public MeasureDefinitionController(
 
     @GetMapping
     @JsonView(Views.Summary.class)
-    public List<MeasureDefinition> getAll() {
+    @PreAuthorize("hasRole('LinkUser')")
```
seanmcilvenna: Does it work to have `@PreAuthorize("hasRole('LinkUser')")` on the entire controller? And then just use `@PreAuthorize("hasAuthority('IsLinkAdmin')")` on the operations that change data?

seanmcilvenna (changes requested):
A couple changes to the measure eval to abstract logic into `shared` module.
Then should update the `validation` service to use the same patterns:
1. Update `Java/validation/pom.xml` 
2. Use security config from `shared`
3. Update controller operations to have `hasRole() and hasAuthorization()`

Java/shared/src/main/java/com/lantanagroup/link/shared/auth/JwtService.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.Map;
+import java.util.function.Function;
+
+import org.springframework.stereotype.Component;
+
+@Component
+public class JwtService {
+  public final long JWT_TOKEN_VALIDITY = 5 * 60 * 60;
+
+  // private static String secret = "/6lBmw6OAVRKb40xHTLy51P6wRrOxlN789c4fUxFB+Dh3wpTTdc2grJuFHr3J6RZyjXoA6E2x83Uv5kHlsljxg==" ;
```
amphillipsLGC: probably want to remove this

Java/shared/src/main/java/com/lantanagroup/link/shared/auth/JwtAuthenticationFilter.java
```diff
@@ -0,0 +1,95 @@
...
+    super();
+    this.secretClient = secretClient;
+    this.jwtService = jwtService;
+    this.handlerExceptionResolver = handlerExceptionResolver;
+  }
+
+  @Override
+  protected void doFilterInternal (HttpServletRequest request, HttpServletResponse response, FilterChain filterChain) throws ServletException, IOException {
+
+    String secret = secretClient.getSecret("link-bearer-key").getValue();
```
amphillipsLGC: does this get called only once? If it will make a request to a vault service, such as key vault, each time it will be pretty slow.

Java/shared/src/main/java/com/lantanagroup/link/shared/auth/JwtService.java
arianamihailescu: Indeed

Java/shared/src/main/java/com/lantanagroup/link/shared/auth/JwtAuthenticationFilter.java
arianamihailescu: I cached it

amphillipsLGC (approved):
Looks good!

PR #236, opened 2024-05-29
LNK-2477: Add link token generation to shared project

amphillipsLGC:
- Added token generation methods and configuration to the shared project
- Fixed swagger in link admin and account services
- Added ability to set symmetric signing key in configuration until long term support for shared encryption across .NET/Java services are enabled


arianamihailescu (approved):
Looks good

PR #237, opened 2024-05-29
LNK-2477: Fixed a bug with the bearer token endpiont not registering

amphillipsLGC:
The bearer token endpoints in link admin were not registering after changes to the config settings to align with updates in the shared project around the link token.


arianamihailescu (approved):
Looks good

PR #246, opened 2024-05-31
LNK-2488 DL and Retry Scenarios - Census

edward-miller-lcg:


DotNet/Census/Listeners/CensusListener.cs
```diff
@@ -89,22 +90,22 @@ private async System.Threading.Tasks.Task StartConsumerLoop(CancellationToken ca
...
+                                    throw new DeadLetterException("Error extracting facility ID and correlation ID: " + ex.Message, AuditEventType.Query, ex);
                                 }
 
                                 deserializedMessage = message;
                                 
                                 try
                                 {
                                     messageMetaData = ExtractFacilityIdAndCorrelationIdFromMessage(rawmessage.Message);
-                                    if (string.IsNullOrWhiteSpace(messageMetaData.facilityId))
+                                    if (string.IsNullOrWhiteSpace(messageMetaData.facilityId) || string.IsNullOrWhiteSpace(messageMetaData.correlationId))
```
dvargaslantana: This looks like it may cause problems. Data Acq does not produce PatientIdsAcquired events with a correlation id

dvargaslantana (approved):

PR #255, opened 2024-06-05
Authentication - Turn off secret client in case of anonymous access s…

arianamihailescu:
…o will not try to connect to KeyVault


amphillipsLGC (approved):
Looks good!

PR #257, opened 2024-06-05
Refactoring configs to use models instead of @Value

seanmcilvenna:


Java/measureeval/src/main/resources/application.yml
```diff
@@ -62,3 +54,10 @@ link:
         path: /api/{facilityId}/QueryResult/{correlationId}
         query:
           - queryType={queryType}
+
+secret-management:
+  key-vault-uri: ''
+
+authentication:
+  anonymous: false
+  authority: ''
```
arianamihailescu, line 63: Those keys (secretManagement:managerUri and authentication:enableAnonymousAccess) were added also in ACA configurations as well as in application.yml with same names. Now they have different names and I do not think if that will work propertly when we switch back and forth between Azure and local based on spring.cloud.azure.appconfiguration.enabled flag. Same for telemetry.exporter.endpoint instead of telemetry.exporterEndpoint

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/configs/SecurityConfig.java
```diff
@@ -12,29 +12,26 @@
 import org.springframework.security.config.annotation.web.builders.HttpSecurity;
 import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
 import org.springframework.security.web.SecurityFilterChain;
-import org.springframework.beans.factory.annotation.Autowired;
 
 
 @Configuration
 @ComponentScan(basePackages = "com.lantanagroup.link.shared.auth")
```
arianamihailescu, line 18: I think we might need to also add com.lantanagroup.link.shared.config in basePackages to find the config files.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/configs/OpenTelemetryConfig.java
```diff
@@ -1,59 +1,70 @@
...
+import io.opentelemetry.sdk.trace.export.BatchSpanProcessor;
 import io.opentelemetry.semconv.ResourceAttributes;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
 
 @Configuration
 public class OpenTelemetryConfig {
```
arianamihailescu, line 27: Do we need to add ComponentScan here  for Config files.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/configs/SecurityConfig.java
seanmcilvenna: We already have `@ConfigurationPropertiesScan("com.lantanagroup.link.shared.config")` in the `MeasureEvalApplication` and the ``ValidationApplicationConfig` classes.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/configs/OpenTelemetryConfig.java
seanmcilvenna: I don't think so... MeasureEval will pick up anything under the `com.lantanagroup.link.measureeval` namespace, which this is under.
But, it brings to mind that this `OpenTelemetryConfig` should probably be moved to `shared` and used by _both_ `measureeval` and `validation`. When _that_ happens, we would need to make sure the `MeasureEvalApplication` and `ValidationApplication` are scanning the shared packages correctly.

Java/measureeval/src/main/resources/application.yml
seanmcilvenna: I'd like to stick to this (hopefully consistent) naming convention... So, when I merge these changes and the builds kick off, I'll update the app config to use the correct properties.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/configs/SecurityConfig.java
arianamihailescu: Are you referring to MeasureEvalApplication class cause I do not see @ConfigurationPropertiesScan("com.lantanagroup.link.shared.config") there

Java/shared/src/main/java/com/lantanagroup/link/shared/auth/SecretClientConfig.java
```diff
@@ -3,25 +3,33 @@
...
+      logger.warn("Key Vault URI is not set. Secret client will not be created.");
+      return null;
+    }
+
     return new SecretClientBuilder()
-            .vaultUrl(vaultUrl)
+            .vaultUrl(this.secretManagementConfig.getKeyVaultUri())
             .credential(new DefaultAzureCredentialBuilder().build())
             .buildClient();
   }
```
arianamihailescu, line 34: Is this not going to require again the az login when Spring tries to construct it? 
return new SecretClientBuilder()
            .vaultUrl(vaultUrl)
            .vaultUrl(this.secretManagementConfig.getKeyVaultUri())
            .credential(new DefaultAzureCredentialBuilder().build())
            .buildClient();

seanmcilvenna: I'm not sure... Would this change affect that? Replacing `vaultUrl` with `this.secretManagementConfig.getKeyVaultUri()`?

arianamihailescu (approved):
Looks good

PR #277, opened 2024-06-11
Data Acquisition - API Guidance Updates (AuthenticationConfigController)

nvmLantana:
Refactors to API to conform with guidelines and expected results.

### DIFF
--- a/DotNet/DataAcquisition/Controllers/AuthenticationConfigController.cs
+++ b/DotNet/DataAcquisition/Controllers/AuthenticationConfigController.cs
@@ -70,7 +70,7 @@ public class AuthenticationConfigController : Controller
     [ProducesResponseType(StatusCodes.Status500InternalServerError)]
     public async Task<ActionResult<AuthenticationConfiguration>> GetAuthenticationSettings(
         string facilityId,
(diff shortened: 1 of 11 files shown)


DotNet/DataAcquisition/Controllers/AuthenticationConfigController.cs
```diff
@@ -73,27 +73,42 @@ private async Task SendAudit(string message, string correlationId, string facili
...
         CancellationToken cancellationToken)
     {
-        if(string.IsNullOrWhiteSpace(facilityId))
-        {
-            return BadRequest();
-        }
-
         try
         {
+            if (queryConfigurationTypePathParameter == null)
```
edward-miller-lcg, line 78: This is always going to be 'false' since it's an enum and it's not defined as nullable. The framework will throw a 500 if it's not provided.

edward-miller-lcg (commented):
Please see comments in review.

DotNet/DataAcquisition/Controllers/AuthenticationConfigController.cs
nvmLantana: Okay - I made it nullable so that we have an opportunity to handle the case and provide information instead of throwing an uncontrolled 500

PR #285, opened 2024-06-13
LNK-2546 Split Query Listener Data Acq

edward-miller-lcg:
Develop generic implementation of a kafka listener that is able to execute custom logic based on the topic.


amphillipsLGC (approved):
Looks good!

PR #291, opened 2024-06-17
Improved link admin service registration logging and link token setting fix

amphillipsLGC:
- Added additional registration logging
- Fixed bug where the link bearer service was not pulling configuration from the updated section name for authority

### DIFF
--- a/DotNet/LinkAdmin.BFF/Infrastructure/Extensions/Security/InitializeSecurity.cs
+++ b/DotNet/LinkAdmin.BFF/Infrastructure/Extensions/Security/InitializeSecurity.cs
@@ -66,7 +66,7 @@ namespace LantanaGroup.Link.LinkAdmin.BFF.Infrastructure.Extensions.Security
 
                 options.Cookie.Name = LinkAdminConstants.AuthenticationSchemes.Cookie;
                 options.Cookie.HttpOnly = configuration.GetValue<bool>("Authentication:Schemas:Cookie:HttpOnly");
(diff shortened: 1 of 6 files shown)


DotNet/LinkAdmin.BFF/Infrastructure/Extensions/Security/InitializeSecurity.cs
```diff
@@ -37,36 +37,63 @@ public static IServiceCollection AddLinkGatewaySecurity(this IServiceCollection
...
                 options.DefaultChallengeScheme = defaultChallengeScheme;
+                
             });
 
             var authBuilder = services.AddAuthentication(options => {
                 options.DefaultScheme = LinkAdminConstants.AuthenticationSchemes.Cookie;
                 options.DefaultChallengeScheme = defaultChallengeScheme;
             });
 
+            logger.Information("Set Default Authentication Scheme: {DefaultScheme}", LinkAdminConstants.AuthenticationSchemes.Cookie);
```
seanmcilvenna, line 60: Feels like a lot of these should be `logger.Debug()` instead of INFO... It doesn't hurt anything to leave it as-is, but consider it...

seanmcilvenna (approved):

DotNet/LinkAdmin.BFF/Infrastructure/Extensions/Security/InitializeSecurity.cs
amphillipsLGC: I updated the logging so the more detailed information is now to to logging level debug while keeping the registration logging information.

PR #299, opened 2024-06-18
Lnk 2061 updates connection validation data acq

edward-miller-lcg:


DotNet/DataAcquisition/Controllers/ConnectionValidationController.cs
```diff
@@ -32,11 +33,14 @@ public ConnectionValidationController(ILogger<ConnectionValidationController> lo
         string facilityId, 
         [FromQuery] string? patientId = default, 
         [FromQuery] string? patientIdentifier = default,
+        [FromQuery] string? measureId = default,
```
seanmcilvenna, line 53: Add new parameters to code-docs above (`<param>`)

seanmcilvenna (approved):

DotNet/DataAcquisition/Controllers/ConnectionValidationController.cs
edward-miller-lcg: done and then some

PR #301, opened 2024-06-20
LNK-2533 Remove GRPC from all services

edward-miller-lcg:


dvargaslantana (approved):
Looks good. Pulled down branch and everything built.

PR #313, opened 2024-06-25
LNK-2580: Fixed the parsing of permissions/roles included in the jwt …

arianamihailescu:
…token to account for those coming as a List or just a String


amphillipsLGC (approved):
looks good!

PR #336, opened 2024-07-09
LNK-2552: Java-services-don't-prioritize-app-config

arianamihailescu:


Java/measureeval/src/main/resources/bootstrap-local.yml
```diff
@@ -0,0 +1,10 @@
+spring:
```
seanmcilvenna, line 1: Don't we want this to be named `bootstrap-dev.yml` so that the DEV environment can use `profile = dev`?

amphillipsLGC (approved):
Looks good!

seanmcilvenna (changes requested):

PR #342, opened 2024-07-10
Set all service versions to 0.1.0 ahead of deployment to staging

sdmcgeown:
Also, make "ServiceInformation" field consistent across the appsettings.json files

### DIFF
--- a/DotNet/Account/Account.csproj
+++ b/DotNet/Account/Account.csproj
@@ -2,6 +2,8 @@
 
   <PropertyGroup>
     <TargetFramework>net8.0</TargetFramework>
+    <Version>0.1.0</Version>
+	  <IncludeSourceRevisionInInformationalVersion>false</IncludeSourceRevisionInInformationalVersion>
 	  <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
     <UserSecretsId>0c1c76b4-6914-4e62-8eb3-c645c905e64b</UserSecretsId>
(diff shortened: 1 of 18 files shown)


seanmcilvenna (approved):
After this is released, do we merge the rc branch into master AND dev so that the version numbers increment for the other branches going forward? 

Also, you might investigate what SNAPSHOTs are in Java/Spring world... I feel like it MAY be suggested that `dev` always use -SNAPSHOT in the version but I'm not sure why and/or what the benefit is.

Regardless, these changes are approved afaic.

DotNet/Account/Account.csproj
amphillipsLGC, line 1: You'll want to add that back as I think this service was one of the ones that were being tested to use the assembly to load the version. Just change the <Version> to 0.1.0.

You can keep the change to the appsettings.json file as it will not break anything.

DotNet/LinkAdmin.BFF/LinkAdmin.BFF.csproj
amphillipsLGC, line 1: You'll want to add that back as I think this service was one of the ones that were being tested to use the assembly to load the version. Just change the to 0.1.0.

You can keep the change to the appsettings.json file as it will not break anything.

amphillipsLGC (changes requested):

DotNet/LinkAdmin.BFF/LinkAdmin.BFF.csproj
sdmcgeown: Hi Adam, I updated the .csproj files as you suggested and fixed indentation issues. Please let me know if the updated PR looks ok and I'll merge it in.

amphillipsLGC (approved):
Looks good!

PR #345, opened 2024-07-12
LNK-2608: DEV - Add azure ad authentication compatibility to link admin bff

amphillipsLGC:
Added Azure Entra ID compatibility in the dev branch

- App registration was created and configured in Azure Entra ID
- Updated the OpenIdConnect auth schema to support Azure Entra Id
- Updated the Cookie auth schema to remove unneeded claims that could cause bloat in the cookie

### DIFF
--- a/DotNet/LinkAdmin.BFF/Infrastructure/Extensions/Security/InitializeSecurity.cs
+++ b/DotNet/LinkAdmin.BFF/Infrastructure/Extensions/Security/InitializeSecurity.cs
@@ -4,6 +4,8 @@ using LantanaGroup.Link.LinkAdmin.BFF.Settings;
 using Link.Authorization.Infrastructure;
 using Link.Authorization.Infrastructure.Extensions;
 using Microsoft.AspNetCore.Authentication;
+using System.IdentityModel.Tokens.Jwt;
+using System.Security.Claims;
 
 namespace LantanaGroup.Link.LinkAdmin.BFF.Infrastructure.Extensions.Security
 {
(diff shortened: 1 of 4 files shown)


DotNet/LinkAdmin.BFF/Infrastructure/Extensions/Security/InitializeSecurity.cs
```diff
@@ -77,7 +79,36 @@ public static IServiceCollection AddLinkGatewaySecurity(this IServiceCollection
...
-                    
+
+                options.Events.OnSigningIn = context =>
+                {
+                    //get claims principal
+                    if (context.Principal?.Identity is not ClaimsIdentity claimsIdentity)
+                    {
+                        return Task.CompletedTask;
+                    }
+                    // Define the claim types to keep
```
seanmcilvenna, line 90: Not necessary for this commit... But is there a reason to make this configurable on the service in the future? So that the appsettings.json file can drive what are `allowedClaims`?

DotNet/LinkAdmin.BFF/Presentation/Endpoints/AuthEndpoints.cs
```diff
@@ -87,8 +89,34 @@ public IResult Login(HttpContext context)
...
-            var user = context.User;
-            return Results.Ok(user.Claims.Select(x => new { x.Type, x.Value }).ToList());
+            //get claims principal
+            if (context.User.Identity is not ClaimsIdentity claimsIdentity)
+            {
+                return Results.Problem("No claims found in the current user identity", statusCode: StatusCodes.Status500InternalServerError);
+            }
+
+            // Define the claim types to keep
+            var allowedClaims = new HashSet<string> {
```
seanmcilvenna, line 99: Consider abstracting `allowedClaims` into something that can be re-used between the two classes.

DotNet/LinkAdmin.BFF/Presentation/Endpoints/AuthEndpoints.cs
```diff
@@ -87,8 +89,34 @@ public IResult Login(HttpContext context)
 
         public IResult GetUser(HttpContext context)
```
seanmcilvenna, line 90: Would be good to have a comment on this public method that provides more details on what this is used for and more importantly what it returns. IResult is very generic and I wouldn't be able to tell at a glance what IResult actually is.

PR #459, opened 2024-09-05
Lnk 2765 as the data acquisition service i want to manage query plans based on the reporting frequency rather than the reporting measure

edward-miller-lcg:
This is only for the API portion. Listener piece to be completed in another ticket.


DotNet/QueryDispatch/Application/Models/ReportableEvents.cs
```diff
@@ -0,0 +1,9 @@
+﻿using LantanaGroup.Link.Shared.Application.Utilities;
+
+namespace QueryDispatch.Application.Models;
+
+public enum ReportableEvents
+{
+    [StringValue("Disharge")]
```
dvargaslantana: This looks misspelled

dvargaslantana (approved):

PR #460, opened 2024-09-06
LNK-2950: Ability to run measure eval using a command line interface

seanmcilvenna:
Adding FileSystemInvocation class that allows evaluating a measure using a data from the locally running file system

Format: `java -jar measureeval.jar -cp com.lantanagroup.link.measureeval.FileSystemInvocation "<measure-bundle-path>" "<patient-bundle-path>" "<start>" "<end>"`

Example: `java -jar measureeval.jar -cp com.lantanagroup.link.measureeval.FileSystemInvocation "C:/path/to/measure-bundle.json" "C:/path/to/patient-bundle.json" "2021-01-01" "2021-12-31"`

Or directly via IntelliJ with a debug configuration.

## Summary by CodeRabbit

- **New Features**
  - Introduced a command-line interface for evaluating healthcare measures using files from the filesystem.
  - Users can specify paths for measure and patient bundles, along with evaluation date ranges.
  - Comprehensive documentation added to guide users on building and running the module.
  - Enhanced flexibility with multiple execution profiles for different use cases.
  - New logging configuration for improved monitoring and debugging.

- **Bug Fixes**
  - Improved error handling during file loading and evaluation processes to provide clearer feedback.


Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/FileSystemInvocation.java
```diff
@@ -0,0 +1,109 @@
...
+import org.hl7.fhir.r4.model.*;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * This class is used to invoke measure evaluation using arbitrary artifacts on the file system for ease of testing and debugging new measures.
+ * Format: `java -jar measureeval.jar -cp com.lantanagroup.link.measureeval.FileSystemInvocation "<measure-bundle-path>" "<patient-bundle-path>" "<start>" "<end>"`
```
smailliwcs: I can't get this syntax to work.  Is the idea here that `measureeval.jar` is a fat JAR produced by `spring-boot:repackage`?  I have to use the following in that situation:

```
java -cp measureeval.jar -Dloader.main=com.lantanagroup.link.measureeval.FileSystemInvocation org.springframework.boot.loader.launch.PropertiesLauncher
```

Because with `java -jar`, the Spring Boot `JarLauncher` will invoke whatever `Start-Class` is baked into the JAR's manifest (which we want to be `MeasureEvalApplication`, not `FileSystemInvocation`).  So you have to explicitly tell Spring Boot to use a different main class via the `loader.main` system property (or the `LOADER_MAIN` environment variable), which `JarLauncher` doesn't respect, but `PropertiesLauncher` does.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/FileSystemInvocation.java
```diff
@@ -0,0 +1,109 @@
...
+                }
+            }
+
+            System.out.println("Loaded " + bundle.getEntry().size() + " resources from directory: " + measureBundlePath);
+
+            return bundle;
+        }
+    }
+
+    public static void main(String[] args) {
```
smailliwcs, line 152: With two main classes now available in the module, we'll want to be explicit about which one is the "true" main class (i.e., the one to launch with `java -jar`).  Because otherwise `spring-boot:repackage` will just pick the one that happens to get compiled first.  Need to add the `mainClass` property to the Spring Boot Maven Plugin's configuration and specify the `MeasureEvalApplication`'s fully qualified class name.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/FileSystemInvocation.java
```diff
@@ -0,0 +1,109 @@
...
+        System.out.println("Loading measure bundle from: " + measureBundlePath);
+
+        File measureBundleFile = new File(measureBundlePath);
+
+        if (!measureBundleFile.exists()) {
+            throw new IllegalArgumentException("Measure bundle file does not exist: " + measureBundlePath);
+        }
+
+        if (measureBundleFile.isFile()) {
+            String measureBundleContent = FileUtils.readFileToString(measureBundleFile);
```
smailliwcs: Recommend using the two-argument version of `readFileToString` so you can tell it which charset to use, otherwise you'll get the VM's default, which won't always be UTF-8.  (Likely to be Windows-1252 on a Windows box.)  Could also consider parsing directly from a `FileInputStream`, rather than copying into a `String` and then parsing that.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/FileSystemInvocation.java
```diff
@@ -0,0 +1,109 @@
...
+            if (measureBundlePath.toLowerCase().endsWith(".json")) {
+                return (Bundle) fhirContext.newJsonParser().parseResource(measureBundleContent);
```
smailliwcs: Recommend using the two-argument version of `parseResource` when a particular concrete resource type is expected.  This avoids the explicit cast and throws a `DataFormatException` if the supplied resource type doesn't match.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/FileSystemInvocation.java
```diff
@@ -0,0 +1,109 @@
...
+ * The `start` and `end` parameters must be in a valid DateTime format from the FHIR specification.
+ * The `measure-bundle-path` can be a path to a single measure bundle file or a directory containing each of the resources needed for the measure.
+ * The `patient-bundle-path` must be a path to a single (JSON or XML) Bundle file or a directory of files containing the patient data to be used in the evaluation.
+ * The response from the operation is the MeasureReport resource being printed to the console in JSON format.
+ */
+public class FileSystemInvocation {
+    private static final FhirContext fhirContext = FhirContext.forR4Cached();
+
+    private static Bundle getBundle(String measureBundlePath) throws IOException {
+        System.out.println("Loading measure bundle from: " + measureBundlePath);
```
smailliwcs: Recommend that all output except for the measure report be directed to stderr.  That way, f you're saving to a file or piping to some other command, you'll _only_ get the measure report in that file/command (and you'll still see messages like this on the console).

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/FileSystemInvocation.java
```diff
@@ -0,0 +1,109 @@
...
+                            System.out.println("Skipping duplicate file: " + filePath);
+                        }
+
+                        Resource resource;
+
+                        if (filePath.endsWith(".json")) {
+                            resource = (Resource) fhirContext.newJsonParser().parseResource(FileUtils.readFileToString(file));
+                        } else {
+                            resource = (Resource) fhirContext.newXmlParser().parseResource(FileUtils.readFileToString(file));
+                        }
```
smailliwcs: Combine with the previous JSON-or-XML parsing logic into a separate method?  (In this case, you don't know the resource type, but you can get away with passing null to the two-argument version of `parseResource` in place of a concrete resource type.  You'll just get back some `IBaseResource` of unknown type.)

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/FileSystemInvocation.java
```diff
@@ -0,0 +1,109 @@
...
+
+            return bundle;
+        }
+    }
+
+    public static void main(String[] args) {
+        String measureBundlePath = args[0];
+        String patientBundlePath = args[1];
+        String start = args[2];
+        String end = args[3];
```
smailliwcs, line 161: Wrap in a try/catch and print usage information to stderr if this fails?

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/FileSystemInvocation.java
```diff
@@ -0,0 +1,109 @@
...
+
+        try {
+            Bundle measureBundle = getBundle(measureBundlePath);
+            Bundle patientBundle = getBundle(patientBundlePath);
+            MeasureEvaluator evaluator = MeasureEvaluator.compile(fhirContext, measureBundle);
+            Patient patient = patientBundle.getEntry().stream()
+                    .filter(e -> e.getResource() instanceof Patient)
+                    .map(e -> (Patient) e.getResource())
+                    .findFirst()
+                    .orElseThrow(() -> new IllegalArgumentException("Patient resource not found in bundle"));
```
smailliwcs: Do we also want an error if there's _more_ than one patient?  If so, consider replacing `findFirst()` with `reduce(StreamUtils::toOnlyElement)`.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/FileSystemInvocation.java
```diff
@@ -0,0 +1,109 @@
...
+            var report = evaluator.evaluate(new DateTimeType(start), new DateTimeType(end),
+                    new StringType("Patient/" + patient.getIdElement().getIdPart()), patientBundle);
+            String json = fhirContext.newJsonParser().encodeResourceToString(report);
+            System.out.println(json);
```
smailliwcs: Again, probably overkill, but could encode directly to `System.out` (by wrapping it in an `OutputStreamWriter`) with `encodeResourceToWriter`.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/FileSystemInvocation.java
```diff
@@ -0,0 +1,109 @@
...
+            }
+        } else {
+            // Parse and load each file in the directory
+            Bundle bundle = new Bundle();
+            bundle.setType(Bundle.BundleType.COLLECTION);
+
+            File[] files = measureBundleFile.listFiles();
+
+            if (files != null) {
+                List<String> loaded = new ArrayList<>();
```
smailliwcs: Make `loaded` a `Set`, since we're mainly using it to call `contains`?

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/FileSystemInvocation.java
```diff
@@ -0,0 +1,109 @@
...
+
+                for (File file : files) {
+                    String filePath = file.getAbsolutePath();
+                    if (file.isFile() && (filePath.endsWith(".json") || filePath.endsWith(".xml"))) {
+                        String fileName = file.getName().substring(0, file.getName().lastIndexOf('.'));
```
smailliwcs: Re: filename wrangling ... not strictly necessary to change, but `FilenameUtils` (Apache Commons) has some relevant methods.  In particular, `getExtension` and `removeExtension` could be helpful/expressive here.  (Note that `getExtension` does not report the period as part of the extension.)

smailliwcs (changes requested):

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/FileSystemInvocation.java
```diff
@@ -0,0 +1,109 @@
...
+ * Format: `java -jar measureeval.jar -cp com.lantanagroup.link.measureeval.FileSystemInvocation "<measure-bundle-path>" "<patient-bundle-path>" "<start>" "<end>"`
+ * Example: `java -jar measureeval.jar -cp com.lantanagroup.link.measureeval.FileSystemInvocation "C:/path/to/measure-bundle.json" "C:/path/to/patient-bundle.json" "2021-01-01" "2021-12-31"`
+ * Careful when specifying file paths to use forward-slash instead of windows' backslash, as that will cause java to interpret the backslash as an escape character and not as a path separator, leading to potentially incorrect parameter interpretation.
+ * The `start` and `end` parameters must be in a valid DateTime format from the FHIR specification.
+ * The `measure-bundle-path` can be a path to a single measure bundle file or a directory containing each of the resources needed for the measure.
+ * The `patient-bundle-path` must be a path to a single (JSON or XML) Bundle file or a directory of files containing the patient data to be used in the evaluation.
+ * The response from the operation is the MeasureReport resource being printed to the console in JSON format.
+ */
```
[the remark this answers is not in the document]
seanmcilvenna: > Adding more details about the expected format for the start and end parameters.

This is already accounted for in "The `start` and `end` parameters must be in a valid DateTime format from the FHIR specification"

> Clarifying the output format and what the user should expect in terms of output content.

This is already accounted for in "The response from the operation is the MeasureReport resource being printed to the console in JSON format."

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/FileSystemInvocation.java
seanmcilvenna: This is fine/good for the ones that we expect to a Bundle, but the other block lower down that deals with `Resource` has to stay as-is because otherwise HAPI complains during runtime. I made the change to the ones casting as a Bundle, though.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/FileSystemInvocation.java
seanmcilvenna: Before you commented on this, I started changing everything to use the logging system. I also modified it to allow for running it against _many_ test samples in a directory... so the output might contain more than one MeasureReport. I'm going to leave as-is, and if we want to consider storing the output in a file, then we can extend it to accept another parameter for "output" or something like that.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/FileSystemInvocation.java
seanmcilvenna: Doing a pre-check of `if args.length != 4` and providing guidance if not.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/FileSystemInvocation.java
seanmcilvenna: Agreed. Changed to use `reduce(StreamUtils::toOnlyElement)`

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/FileSystemInvocation.java
seanmcilvenna: Using logging system instead. May consider output parameter in future.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/FileSystemInvocation.java
seanmcilvenna: Switched to a HashSet

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/FileSystemInvocation.java
seanmcilvenna: Made some minor mods, if for no other reason than to get more familiar with `FilenameUtils`

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/FileSystemInvocation.java
seanmcilvenna: Good point. I missed this because I was only running it from within IntelliJ. I updated docs to use your sytnax above, and tested that it works from the command line after building the jar with `mvn install -pl measureeval -am`.

seanmcilvenna:
@smailliwcs In this round of changes I added a `logback.xml` file to support logging when running as a CLI... It doesn't _seem_ to be conflicting with spring-boot... But, what do you think? Is that a problem?

smailliwcs:
Pushed a commit to allow choosing (at build time via a Maven profile) whether to use the API (default) or the CLI as the main class.  This allows you to run the CLI via `java -jar` without the `loader.main`/`PropertiesLauncher` rigmarole.  Updated the README to reflect this change.

> @smailliwcs In this round of changes I added a `logback.xml` file to support logging when running as a CLI... It doesn't _seem_ to be conflicting with spring-boot... But, what do you think? Is that a problem?

In my testing, `logback.xml` overrides `logback-spring.xml`, so I don't think that will work.  One option would be to specify a CLI-specific configuration file at invocation time:

```
java -Dlogback.configurationFile=/path/to/logback-cli.xml -jar measureeval.jar ...
```

smailliwcs (approved):

PR #466, opened 2024-09-12
LNK-2926 Add anonymous access check to shared tenant check

edward-miller-lcg:


DotNet/Tenant/Services/FacilityConfigurationService.cs
```diff
@@ -316,12 +321,12 @@ private async Task MeasureDefinitionExists(String reportType)
...
 
-                if (_linkTokenServiceConfig.Value.SigningKey is null)
+                //TODO: add method to get key that includes looking at redis for future use case
+                if (!_linkBearerServiceOptions.Value.AllowAnonymous && _linkTokenServiceConfig.Value.SigningKey is null)
                     throw new Exception("Link Token Service Signing Key is missing.");
 
-
                 //get link token
-                if (_linkTokenServiceConfig.Value.EnableTokenGenerationEndpoint)
+                if (!_linkBearerServiceOptions.Value.AllowAnonymous && _linkTokenServiceConfig.Value.EnableTokenGenerationEndpoint)
```
arianamihailescu: Do we need the AllowAnonymous check here? I thought that the AllowAnonymous flag is for calling into Tenant service with a token not calling from Tenant service with a token.

edward-miller-lcg: My understanding is that it would be across the board for incoming and outgoing requests. If this isn't there and it tries to get a token when nothing is generating it, then we get errors. Adding @amphillipsLGC for this discussion.

amphillipsLGC: For now just leave it there. I'll make an architecture ticket to review how we want to do this because there might be a cleaner way to go about this. This is to mainly get the local compose working again.

amphillipsLGC: After discussing this it was decided to remove the `_linkTokenServiceConfig.Value.EnableTokenGenerationEndpoint` flag from the if statement.

dvargaslantana (approved):

PR #474, opened 2024-09-18
LNK-2768 dataacq reportable event field

edward-miller-lcg:


DotNet/DataAcquisition/Application/Services/ValidateFacilityConnectionService.cs
dvargaslantana, line 1: Does 'Frequency' need to be added?

dvargaslantana (approved):

PR #548, opened 2024-11-13
LNK-3172: fix-put-query-list

edward-miller-lcg:
### 🛠 Description of Changes

Add pull of existing entity and copying new values in properties to existing entity and then save.

### 🧪 Testing Performed

local testing.


amphillipsLGC (approved):
looks good!

PR #562, opened 2024-12-05
LNK-3212: Address security cookie dynamic scan result - Link Admin

amphillipsLGC:
### 🛠 Description of Changes

- Removed the expiration setting for the `link_cookie` to ensure it will be a session cookie.
- Made the backend service health check configurable, disabled by default

### 🧪 Testing Performed

- Logged into link-admin
- Made a request to the Account service to ensure everything was working correctly
- Closed my browsing session
- Verified that the cookie was no longer present and received an unauthorized response when making the same request to the Account service.


arianamihailescu (approved):
Looks good

PR #572, opened 2024-12-12
TECH_DEBT: Measure debuggability

seanmcilvenna:
### 🛠️ Description of Changes

This is considered TECH DEBT because the core/underlying functionality is not meaningfully modified. The changes here are intended to provide better documentation and better support for debugging when measure evaluation doesn't return the expected results.

- Introduced a new endpoint to retrieve CQL for measure definitions for testing purposes.
- Added a `cqlDebug` configuration property for enhanced debugging.
- New logging appender for CQL evaluations to improve logging capabilities.
- Streamlined method signatures for evaluating patient bundles.
- Enhanced error handling and parameter management in evaluation methods.
- Updated documentation to include new configuration details for Kafka and Azure App Config.
- Modified CD pipeline to build and publish the CLI JAR
- Added all Azure Pipelines to a solution folder in the VS SLN so that they can be easily opened from an IDE

### 🧪 Testing Performed

Ran the measureeval service locally and used the REST API to invoke $evaluate to ensure the evaluator still functioned correctly.
Ran the FileSystemInvocation class from the command line to test evaluation against other measures.

### 📓 Documentation Updated

Updated the measureeval service spec document to bring it up-to-date; it looks like it was already inaccurately describing the service's configuration. Added notes about the additional "cql debugging" functionality that's now available.

## Summary by CodeRabbit

Based on the comprehensive changes, here are the updated release notes:

- **New Features**
  - Added CQL (Clinical Quality Language) debugging capabilities.
  - Introduced a new API endpoint for retrieving measure library CQL.
  - Enhanced measure evaluation with an optional debug mode.
  - Improved logging and tracing for the measure evaluation process.

- **Improvements**
  - Streamlined measure evaluation method signatures.
  - Updated FHIR bundle transaction processing.
  - Enhanced library and CQL retrieval utilities.
  - Refined configuration management for measure evaluation.

- **Configuration Changes**
  - Added new configuration options for CQL debugging.
  - Updated logging configurations.
  - Introduced new environment variables for service configuration.

- **Documentation**
  - Updated service specifications and README with new configuration details.
  - Expanded documentation for measure evaluation and testing.

These changes focus on improving the measure evaluation service's flexibility, debugging capabilities, and overall configuration management.


Java/measureeval/src/main/resources/logback-cli.xml
```diff
@@ -7,16 +7,14 @@
...
     </appender>
+    <appender name="cql" class="com.lantanagroup.link.measureeval.utils.CqlLogAppender"/>
 
-    <logger name="com.lantanagroup.link" level="info" additivity="false">
-        <appender-ref ref="CONSOLE"/>
-    </logger>
-
-    <logger name="org.opencds.cqf" level="debug" additivity="false">
-        <appender-ref ref="CONSOLE"/>
+    <logger name="com.lantanagroup.link" level="info"/>
```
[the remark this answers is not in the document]
seanmcilvenna: Considered. Ignoring.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/utils/CqlUtils.java
```diff
@@ -0,0 +1,80 @@
...
+                rangeCql.append(lines[i].substring(0, endColumn));
+            } else {
+                rangeCql.append(lines[i]);
+            }
+            if (i != endLine - 1) {
+                rangeCql.append("\n");
+            }
+        }
+        return rangeCql;
+    }
```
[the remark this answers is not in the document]
seanmcilvenna: Implemented/addressed.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/services/MeasureEvaluatorCache.java
```diff
@@ -25,7 +28,7 @@ public MeasureEvaluator get(String id) {
             if (measureDefinition == null) {
                 return null;
             }
-            return MeasureEvaluator.compile(fhirContext, measureDefinition.getBundle());
+            return MeasureEvaluator.compile(fhirContext, measureDefinition.getBundle(), this.linkConfig.isCqlDebug());
```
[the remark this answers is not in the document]
seanmcilvenna: Considered. Ignoring, for now.

docs/service_specs/measure_eval.md
```diff
@@ -11,26 +11,68 @@ The Measure Eval service is a Java based application that is primarily responsib
...
+| link.cql_debug               | false                                                                 | No      |
+
+### Database (Mongo)
+
+| Name                             | Value       | Secret? |
+|----------------------------------|-------------|---------|
+| spring.data.mongodb.host         |             | No      |
+| spring.data.mongodb.port         |             | No      |
+| spring.data.mongodb.database     | measureeval | No      |
+| TODO: Add authentication details |             | No      |
```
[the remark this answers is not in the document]
seanmcilvenna: Going to address TODO in a future PR

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/services/MeasureEvaluator.java
```diff
@@ -100,38 +107,50 @@ private MeasureReport doEvaluate(
...
+        // Output debug/trace information about the results of the evaluation
         if (logger.isTraceEnabled()) {
+            // Output the group/population counts
+            for (MeasureReport.MeasureReportGroupComponent group : doEvaluate(periodStart, periodEnd, subject, additionalData).getGroup()) {
+                logger.trace("Group {}: {}", group.getId(), group.getPopulation().size());
+                for (MeasureReport.MeasureReportGroupPopulationComponent population : group.getPopulation()) {
+                    logger.trace("Population {}: {} - {}", population.getCode().getCodingFirstRep().getDisplay(), population.getCount());
+                }
+            }
+
```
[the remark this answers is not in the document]
seanmcilvenna: This should not be part of this tech debt PR.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/repositories/LinkInMemoryFhirRepository.java
```diff
@@ -0,0 +1,36 @@
...
+                if (!entry.getResource().hasId()) {
+                    entry.getResource().setId(java.util.UUID.randomUUID().toString());
+                }
+
+                this.update(entry.getResource());
+            }
+        }
+
+        return transaction;
+    }
```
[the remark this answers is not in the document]
seanmcilvenna: Addressed.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/utils/CqlLogAppender.java
```diff
@@ -0,0 +1,93 @@
...
+                groupedOutput.append(",");
+            }
+            groupedOutput.append(entry.getKey());
+            if (entry.getValue() > 1) {
+                groupedOutput.append("(").append(entry.getValue()).append(")");
+            }
+        }
+
+        return groupedOutput.toString();
+    }
```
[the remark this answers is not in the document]
seanmcilvenna: The log already has all these resources enumerated. The grouping will only ever be smaller than the log would otherwise be.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/utils/CqlLogAppender.java
```diff
@@ -0,0 +1,93 @@
...
+            String output = matcher.group(3)
+                    .replaceAll("org.hl7.fhir.r4.model.", "")
+                    .replaceAll("@[0-9A-Fa-f]{6,8}", "");
+            String cql = null;
+
+            // Group the resources in the output
+            output = groupResources(output);
+
+            MeasureEvaluatorCache measureEvalCache = ApplicationContextProvider.getApplicationContext().getBean(MeasureEvaluatorCache.class);
+            MeasureEvaluator evaluator = measureEvalCache.get(libraryId);       // Assume the measure id is the same as the library id since the log entry doesn't output the measure url/id
```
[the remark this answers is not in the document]
seanmcilvenna: This is an slf4j class that can't be dependency injected using Spring. ApplicationContextProvider was built for this sole purpose that the log appender can't be dependency-injected, yet it needs the context of the application to get access to other services.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/repositories/LinkInMemoryFhirRepository.java
smailliwcs, line 1: Why do we need this?  Just to preprocess each resource to ensure that it has an ID?  But if we receive a bundle where that's not the case, isn't it an upstream issue?  (Something wrong with the tooling that produces the measure definition bundles?)  Or at least something that we could resolve in a separate preprocessing step that's not integrated with `InMemoryFhirRepository`?  Alternatively, if this is the best place to inject that logic, why not do all of the preprocessing up front (i.e., for all entries) and then delegate to `super.transaction`?  (Though with that strategy, we'd need to check each entry's `request.method` and, if it's a `PUT`, tweak the `request.url` to match the generated ID.)

I just worry that we're bypassing a lot of logic in the superclass, which IIUC is not exactly stable.  Not to mention that we're changing the meaning of what we've overridden: instead of "transact the provided bundle and return a response bundle with the outcomes," it's "treat everything like a `PUT` and return the bundle as is."  Which, given our use case, _probably_ won't come back to bite us, but still ...

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/services/MeasureEvaluator.java
```diff
@@ -117,38 +124,50 @@ private MeasureReport doEvaluate(
...
                 "Evaluating measure: MEASURE=[{}] START=[{}] END=[{}] SUBJECT=[{}] RESOURCES=[{}]",
                 measure.getUrl(), periodStart.asStringValue(), periodEnd.asStringValue(), subject, entries.size());
+
+        // Output debug/trace information about the results of the evaluation
         if (logger.isTraceEnabled()) {
+            // Output the group/population counts
+            for (MeasureReport.MeasureReportGroupComponent group : doEvaluate(periodStart, periodEnd, subject, additionalData).getGroup()) {
+                logger.trace("Group {}: {}", group.getId(), group.getPopulation().size());
+                for (MeasureReport.MeasureReportGroupPopulationComponent population : group.getPopulation()) {
+                    logger.trace("Population {}: {} - {}", population.getCode().getCodingFirstRep().getDisplay(), population.getCount());
```
smailliwcs: Missing an argument?  There are three placeholders but only two values to inject.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/configs/LinkConfig.java
```diff
@@ -16,6 +16,8 @@
 public class LinkConfig {
     private String reportabilityPredicate;
 
+    private boolean cqlDebug = false;
```
smailliwcs, line 19: Is there still a use case for this to be set to true?  Give the volume of logging that this produces and the performance impact of trawling the CQL for each log message (or maybe that's not an issue?), do we want to restrict debugging to either the CLI or explicit calls to `POST /api/measure-definition/{id}/$evaluate`?

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/utils/CqlLogAppender.java
```diff
@@ -0,0 +1,93 @@
...
+import com.lantanagroup.link.measureeval.services.MeasureEvaluatorCache;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class CqlLogAppender extends AppenderBase<ILoggingEvent> {
```
smailliwcs, line 18: Not critical, but since this needs the Spring application context, why not make it an `ApplicationContextAware` `@Component`?  Rather than statically calling into the new `ApplicationContextProvider` (which could then be removed).  Although with this strategy we'd need to add the appender to the Logback context programmatically, rather than specifying it in the XML config.  This is basically the "simple approach" described here:

https://stackoverflow.com/a/55575749/24515610

And on second thought, we _don't_ need the whole application context ... we just need a single bean.  Which we could wire in directly (as a constructor argument) if this class were a `@Component`.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/utils/CqlLogAppender.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        return groupedOutput.toString();
+    }
+
+    private void processLogEntry(String libraryId, String range, String output, String cql) {
+        if (cql != null) {
+            Pattern definePattern = Pattern.compile("^define \"([^\"]+)\"");
+            Matcher matcher = cql != null ? definePattern.matcher(cql) : null;
+            if (matcher != null && matcher.find()) {
```
smailliwcs: Given the containing `if`, neither `cql` nor `matcher` can be null in this block.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/utils/CqlUtils.java
```diff
@@ -0,0 +1,84 @@
...
+import org.hl7.fhir.r4.model.Bundle;
+import org.hl7.fhir.r4.model.Library;
+import org.hl7.fhir.r4.model.ResourceType;
+import org.springframework.http.HttpStatus;
+import org.springframework.web.server.ResponseStatusException;
+
+import java.util.Optional;
+
+public class CqlUtils {
+    public static String getCql(Bundle bundle, String libraryId, String range) {
```
smailliwcs: Do we want to throw `ResponseStatusException` from this method?  It's appropriate when called from the new `$cql` endpoint, but not from `CqlLogAppender` (which would bring down a measure evaluation, maybe halfway through handling a Kafka event).  In the latter case, we'd probably want to log a warning or just fail silently.

Java/measureeval/src/main/resources/logback-spring.xml
```diff
@@ -41,17 +42,20 @@
         </then>
     </if>
 
+    <logger name="org.opencds.cqf.cql.engine.debug.DebugUtilities" level="DEBUG" additivity="false">
+        <appender-ref ref="cql"/>
+    </logger>
+    <logger name="com.lantanagroup.link.measureeval.services" level="WARN"/>
```
smailliwcs: Tangential, but do we have a project convention on logging levels?  I gravitate toward using a root level of `INFO` and then cranking down the entirety of `com.lantanagroup.link` to `DEBUG`.  (Regardless, if we set the root level to `WARN`, there's no need to specify `WARN` for individual loggers.)

smailliwcs (changes requested):

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/repositories/LinkInMemoryFhirRepository.java
seanmcilvenna: This implementation is primarily used to avoid the exception stack trace that is thrown when the InMemoryFhirRepository.transaction method is called by measure eval, but is not implemented in the default InMemoryFhirRepository. The exception and stack trace are unnecessarily noisy in the logs and when I was trying to debug cql using the logs, I felt disrupted by it.

seanmcilvenna: What super logic are you referring to? I don't see any super logic that this class overrides... All I see is this:

```
    public <B extends IBaseBundle> B transaction(B transaction, Map<String, String> headers) {
        throw new NotImplementedOperationException("The transaction operation is not currently supported");
    }
```

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/configs/LinkConfig.java
seanmcilvenna: I think that's a good consideration in the future... But, I'd like to keep `cqlDebug` configuration _for now_ until we have a way to get the exact patient bundle that's used for evaluation out of the system that _could_ be used with `/$evaluate`... As far as I know, there's no way to ask the system to store the patient bundles that are input to the measure eval, so that we could grab those and test against them with `/$evaluate`.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/repositories/LinkInMemoryFhirRepository.java
smailliwcs: Looks like it's called `transactionStub` in the version of CQF we're using ... but that logic was moved into `transaction` in more recent versions.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/utils/CqlLogAppender.java
seanmcilvenna: I was under the impression that `AppenderBase` is not instantiated using Spring, so attributing it with `@Component` wouldn't really do anything. I see this in the link you provided:

> It isn't possible to do what you are trying to do. Logback is initialised before the application context is created so there's nothing to perform the dependency injection.

And when I test it with `@Component` I get the following exception:

```
09:36:03,547 |-ERROR in ch.qos.logback.core.model.processor.AppenderModelHandler - Could not create an Appender of type [com.lantanagroup.link.measureeval.utils.CqlLogAppender]. ch.qos.logback.core.util.DynamicClassLoadingException: Failed to instantiate type com.lantanagroup.link.measureeval.utils.CqlLogAppender
	at ch.qos.logback.core.util.DynamicClassLoadingException: Failed to instantiate type com.lantanagroup.link.measureeval.utils.CqlLogAppender
```

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/utils/CqlUtils.java
seanmcilvenna: Changed to `NotFoundException` in `CqlUtils` and handled that exception in its usages.

Java/measureeval/src/main/resources/logback-spring.xml
seanmcilvenna: Agreed. Changed root level to INFO and link-specific classes to WARN.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/utils/CqlLogAppender.java
smailliwcs: Hmm, tried to test this out myself, and I can't get _any_ CQL debugging output from the CLI.  Which I guess makes sense ... the `CqlLogAppender` ultimately relies on the Spring application context to get the measure definition bundle, but the CLI isn't a Spring application, so there _is_ no application context.  (This results in `append` throwing an NPE, which Logback swallows.)  So maybe a more important issue (and bigger refactor) than I originally thought.

seanmcilvenna: OK to resolve this conversation, then, and follow up more on it later?

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/repositories/LinkInMemoryFhirRepository.java
seanmcilvenna: I'm happy removing `LinkInMemoryFhirRepository` after we upgrade and get to take advantage of those refactors to `InMemoryFhirRepository`.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/utils/CqlLogAppender.java
smailliwcs: Fine by me if we're OK without CQL logging in the CLI.  But if that's the case, we should probably remove `CqlLogAppender` from the CLI's Logback config.

seanmcilvenna: Well... No. Not OK without CQL logging in the CLI. But, CQL logging in CLI works for me, using the currently implemented use of `ApplicationContextProvider`. If I replace that with something that uses @Component, _then_ I don't get any CQL logging in the CLI.

smailliwcs: Hmm, I must be running the CLI differently.  Are you running as described in the README (`mvn package` followed by `java -jar`)?  When I do that, I don't see any CQL logging.  And when I step through in IntelliJ using a run/debug configuration that references logback-cli.xml as the `logback.configurationFile` JVM parameter, I can see that `ApplicationContextProvider.getApplicationContext()` returns null.

seanmcilvenna:
I'm gonna have to test again. I tested against the CLI early on... Maybe I broke something in the process.

sdmcgeown (approved):
Looks good. Was auto-added as a reviewer because of the change to pom.xml where version numbers can potentially be changed.

smailliwcs (approved):

PR #597, opened 2024-12-31
LNK-3310: Admin UI doesn't work when anonymous authentication is configured

seanmcilvenna:
### 🛠️ Description of Changes

Always configuring CORS in BFF regardless of anonymous auth config.
Detecting if login is required in AppComponent.
Showing navigation when login is _not_ required.
Adding extra logging to CORS configuration/initialization for BFF.
Extra: Fixing docker compose healthcheck command on MSSQL.

### 🧪 Testing Performed

Built/ran BFF in Docker Compose, which is configured to use anonymous authentication. Observed that CORS was correctly being registered regardless of anonymous auth.
Built/ran UI in both Docker Compose and locally against the BFF running in Docker Compose.
Opened the UI and observed that it was correctly showing the navigation, and that the Tenants screen successfully received responses from the BFF for a list of tenants (although I don't have any tenants configured). Observed similar successes when looking at audits (though I don't have any audits in the db).

### 📓 Documentation Updated

- Updated README.md to reflect new project name.


sdmcgeown (approved):
Looks good

amphillipsLGC (approved):
Looks good to me!

PR #764, opened 2025-03-31
LNK-3611: Versioning System

nvmLantana:
### 🛠️ Description of Changes

Added VersionInfo.json to measureEval and validation services, with settings to copy them to the output directory in pom.xml.

In .NET, added to shared with settings to copy to output directory on build, added a Shared VersionInfo class to load and return the file.

This is so Stephen can set service-specific naming and versioning when he does a release.

### 🧪 Testing Performed
I made Unit Tests in .NET to test the file is output and readable.

### 🧑‍🔬 Unit Testing

### 📓 Documentation Updated

## Summary by CodeRabbit

- **New Features**
  - The application now retrieves version information dynamically, displaying the current product name and version consistently from a dedicated configuration file.

- **Chores**
  - Project configurations and dependency references have been streamlined to improve maintainability and overall reliability.


Java/measureeval/src/main/resources/VersionInfo.json
```diff
@@ -0,0 +1,4 @@
+{
+  "VersionName": "NHSNLink",
+  "VersionNumber": "0.1.1"
+}
```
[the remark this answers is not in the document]
nvmLantana: No. Each microservice will have its own version and name, so they have their own file.

Java/measureeval/pom.xml
```diff
@@ -161,6 +161,14 @@
...
 
     <build>
+        <resources>
+            <resource>
+                <directory>src/main/resources</directory>
+                <includes>
+                    <include>VersionInfo.json</include>
+                </includes>
+            </resource>
+        </resources>
```
smailliwcs: Not necessary to add this.  By default, everything in `src/main/resources` is included.  By specifying an explicit `include`, it effectively _excludes_ everything else, which we don't want.

Java/measureeval/src/main/resources/VersionInfo.json
smailliwcs, line 1: The version number is already available in the POM itself.  Worried that duplicating it here will end up with one or the other copy getting out of sync.  Consider removing this file and instead relying on `spring-boot:build-info`, which we're already using in Measure Eval, and which produces a `META-INF/build-info.properties` file in the output directory that includes the version (among other metadata).  But need to confirm with Stephen that that process/format would work.

Java/validation/pom.xml
```diff
@@ -149,6 +149,14 @@
...
 
     <build>
+        <resources>
+            <resource>
+                <directory>src/main/resources</directory>
+                <includes>
+                    <include>VersionInfo.json</include>
+                </includes>
+            </resource>
+        </resources>
```
smailliwcs: See previous comments.  Remove this and instead configure the Spring Boot Maven Plugin to produce build metadata in the output directory?

```xml
<plugin>
    <groupId>org.springframework.boot</groupId>
    <artifactId>spring-boot-maven-plugin</artifactId>
    <executions>
        <execution>
            <goals>
                <goal>build-info</goal>
                <goal>repackage</goal>
            </goals>
        </execution>
    </executions>
    <configuration>
        <excludes>
            <exclude>
                <groupId>org.projectlombok</groupId>
                <artifactId>lombok</artifactId>
            </exclude>
        </excludes>
    </configuration>
</plugin>
```

link-cloud.sln
```diff
@@ -19,6 +19,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 	ProjectSection(SolutionItems) = preProject
 		.editorconfig = .editorconfig
 		README.md = README.md
+		version.json = version.json
```
smailliwcs: Not seeing this file in Git.  Was it supposed to have been added?

smailliwcs (changes requested):

Java/measureeval/pom.xml
nvmLantana: AI screwed me on this one. Will fix.

link-cloud.sln
nvmLantana: Nope! It isn't. I missed deleting a reference to the file in a csproj file from an earlier iteration

Java/measureeval/src/main/resources/VersionInfo.json
nvmLantana: Is that outputted file accessible/able to be pulled into the code if it needs to be referenced, perhaps by an API GetVersion call or something?

smailliwcs: Yes, easiest way is to use the `BuildProperties` bean that Spring Boot auto-configures if that file is found.  Which you can use to `getVersion` and such.

https://docs.spring.io/spring-boot/api/java/org/springframework/boot/info/BuildProperties.html

seanmcilvenna (changes requested):
Please provide some basic guidance in the `/docs/config` files on how to specify the version information

PR #770, opened 2025-04-04
LNK-3608: auth token for the fhir server

edward-miller-lcg:
### 🛠️ Description of Changes
Add caching strategy for fhir api calls for the access token to be retrieved from EHRs identity server. 

### 🧪 Testing Performed
local

### 🧑‍🔬 Unit Testing
- [X ] I have written or updated unit tests to cover my changes

### 📓 Documentation Updated
n/a

## Summary by CodeRabbit

- **New Features**
  - Introduced facility-specific authentication, allowing for more contextual security.
  - Integrated flexible caching options with configurable expiration strategies to boost performance.

- **Bug Fixes**
  - Improved error handling and validation for authentication processes.

- **Refactor**
  - Streamlined authentication and API interfaces to ensure consistent parameter handling.
  - Reorganized internal logic for enhanced error checking and maintainability.

### DIFF
--- a/DotNet/DataAcquisition/Application/Models/CacheSettings.cs
+++ b/DotNet/DataAcquisition/Application/Models/CacheSettings.cs
@@ -1,6 +1,0 @@
-﻿namespace LantanaGroup.Link.DataAcquisition.Application.Models;
-
-public class CacheSettings
-{
-    public double CacheTimeoutInMinutes { get; set; } = 20;
-}
--- a/DotNet/DataAcquisition/Application/Models/CacheSettings.cs
+++ b/DotNet/DataAcquisition/Application/Models/CacheSettings.cs
@@ -1,0 +1,6 @@
+﻿namespace LantanaGroup.Link.DataAcquisition.Application.Models;
+
+public class CacheSettings
+{
+    public double CacheTimeoutInMinutes { get; set; } = 20;
+}
--- a/DotNet/DataAcquisition/Application/Services/Auth/EpicAuth.cs
+++ b/DotNet/DataAcquisition/Application/Services/Auth/EpicAuth.cs
@@ -1,5 +1,7 @@
-﻿using LantanaGroup.Link.DataAcquisition.Domain.Models;
+﻿using LantanaGroup.Link.DataAcquisition.Application.Models;
+using LantanaGroup.Link.DataAcquisition.Domain.Models;
 using LantanaGroup.Link.DataAcquisition.Services.Interfaces;
+using LantanaGroup.Link.Shared.Application.Extensions.Caching;
 using Microsoft.IdentityModel.JsonWebTokens;
 using Microsoft.IdentityModel.Tokens;
 using Org.BouncyCastle.Crypto;
@@ -16,11 +18,15 @@ public class EpicAuth : IAuth
 {
     private readonly HttpClient _httpClient;
     private readonly ILogger<EpicAuth> _logger;
+    private readonly InMemoryCacheService _cacheService;
+    private readonly CacheSettings _cacheSettings;
 
-    public EpicAuth(HttpClient httpClient, ILogger<EpicAuth> logger)
+    public EpicAuth(HttpClient httpClient, ILogger<EpicAuth> logger, InMemoryCacheService cacheService, CacheSettings cacheSettings)
     {
         _httpClient = httpClient;
         _logger = logger;
+        _cacheService = cacheService;
+        _cacheSettings = cacheSettings;
     }
 
     /// <summary>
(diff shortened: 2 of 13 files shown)


DotNet/DataAcquisition/Application/Models/CacheSettings.cs
smailliwcs, line 1: A problem with this approach is that the EHR dictates the appropriate cache timeout, not Link.  Typically the token response includes an `expires_in` property for this purpose ... see [Epic's docs](https://fhir.epic.com/Documentation?docId=oauth2&section=Backend-Oauth2_Access-Token-Request).  In some cases, the timeout can be quite short (9.5 minutes for Cerner).  The approach we took in MVP was to just hammer on the token until we get an HTTP 401 (which triggers a refresh), though you could certainly use `expires_in` (as long as it's provided) to trigger a refresh proactively.

DotNet/DataAcquisition/Application/Services/Auth/EpicAuth.cs
```diff
@@ -29,8 +35,13 @@ public EpicAuth(HttpClient httpClient, ILogger<EpicAuth> logger)
     /// <param name="httpClient"></param>
     /// <param name="authSettings"></param>
     /// <exception cref="NotImplementedException"></exception>
-    public async Task<(bool isQueryParam, object authHeaderValue)> SetAuthentication(AuthenticationConfiguration authSettings)
+    public async Task<(bool isQueryParam, object authHeaderValue)> SetAuthentication(string facilityId, AuthenticationConfiguration authSettings)
```
smailliwcs, line 45: Is it possible that we'll hit this logic simultaneously from multiple threads?  If so, synchronization may be needed to ensure we don't request a token separately on each thread.  (Or maybe better to put that synchronization in `AuthMessageHandlerFactory.Build` so we don't have to think about it within each `IAuth` implementation.)

edward-miller-lcg: Right now, no. We don't do any multi-threading within data acquisition at the moment, though, that is an enhancement that needs to be implemented. I think kicking the can down the road is appropriate here. I'll mention that to Adam though.

DotNet/DataAcquisition/Application/Models/CacheSettings.cs
edward-miller-lcg: Thanks for pointing this out. I added code to handle this.

smailliwcs, line 1: With the change to using `expires_in` to determine each cache entry's expiration, this class is no longer needed?

DotNet/DataAcquisition/Application/Services/Auth/EpicAuth.cs
```diff
@@ -46,9 +66,12 @@ public EpicAuth(HttpClient httpClient, ILogger<EpicAuth> logger)
 
             if (responseJson != null)
             {
+                var expirationInSeconds = responseJson.RootElement.GetProperty("expires_in").GetInt32();
                 var accessToken = Sanitize(responseJson.RootElement.GetProperty("access_token").GetString());
                 if (!string.IsNullOrWhiteSpace(accessToken))
                 {
+                    _cacheService.Set(facilityId, accessToken, TimeSpan.FromSeconds(expirationInSeconds));
```
smailliwcs: `InMemoryCacheService.Set` assumes a sliding expiration, but we need an absolute expiration.

DotNet/DataAcquisition/appsettings.json
```diff
@@ -1,4 +1,7 @@
-{  
+{
+  "Cache": {
+    "CacheTimeoutInMinutes": 59
+  },
```
smailliwcs: No longer needed if removing `CacheSettings`.

smailliwcs (changes requested):

DotNet/DataAcquisition/Application/Services/Auth/EpicAuth.cs
edward-miller-lcg: made update to accomodate this.

DotNet/DataAcquisition/appsettings.json
edward-miller-lcg: removed

DotNet/DataAcquisition/appsettings.json
```diff
@@ -1,4 +1,4 @@
-{  
+{S
```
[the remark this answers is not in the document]
edward-miller-lcg: fixed

DotNet/DataAcquisition/Application/Models/CacheSettings.cs
edward-miller-lcg: removed

smailliwcs (approved):

PR #853, opened 2025-05-27
LNK-3787: Report service logging improvements

amphillipsLGC:
### 🛠️ Description of Changes
Updated logging for dead letter and transient exception handlers.

### 🧪 Testing Performed
Produced messages that resulted in various dead letter and transient exceptions and verified that needed information was present in the logs.

### 🧑‍🔬 Unit Testing

### 📓 Documentation Updated

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved logging details and clarity for error and exception events, including more structured and informative log messages.
  - Enhanced exception handling to provide more accurate context in error logs, including timestamps, subscribed topics, and message offsets.
- **Refactor**
  - Removed unused code and streamlined variable declarations for better code maintainability.
  - Updated method signatures to remove unnecessary null checks, improving exception handling flow.
- **Style**
  - Standardized logging practices with named placeholders for increased readability and consistency.


DotNet/Shared/Application/Error/Interfaces/ITransientExceptionHandler.cs
```diff
@@ -17,7 +17,7 @@ public interface ITransientExceptionHandler<K, V>
         /// </summary>
         public string ServiceName { get; set; }
 
-        void HandleException(ConsumeResult<K, V> consumeResult, string facilityId, string message = "");
+        void HandleException(ConsumeResult<K, V>? consumeResult, string facilityId, string message = "");
```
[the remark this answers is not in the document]
amphillipsLGC: Based on your feedback, I reverted the change to not allow a null consumeResult, and instead removed the null check from the method that would always return false. Does this seem like a valid solution?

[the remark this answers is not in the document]
amphillipsLGC: I believe I removed the remaining mismatch. Did I miss any others?

DotNet/Report/Listeners/ResourceEvaluatedListener.cs
```diff
@@ -238,8 +241,8 @@ await resourceManager.UpdateResourceAsync(existingReportResource,
                             }
                             catch (TimeoutException ex)
                             {
+                                _logger.LogError(ex, "Timeout exception encountered on {date} for topics: [{consumers}] at offset: {offset}", DateTime.UtcNow, string.Join(", ", consumer.Subscription), result.TopicPartitionOffset);
                                 var transientException = new TransientException(ex.Message, ex.InnerException);
```
smailliwcs: Isn't this problematic?  We're basically throwing away the `TimeoutException` here.  And if it doesn't have an `InnerException`, all we're left with is the `Message` that we're propagating to the `TransientException`.  Don't we want to pass along `ex` as is?  (Could use the three-arg overload for this?)

DotNet/Shared/Application/Error/Handlers/DeadLetterExceptionHandler.cs
```diff
@@ -32,33 +31,34 @@ public void HandleException(ConsumeResult<K, V> consumeResult, string facilityId
...
             {
-                Logger.LogError(e, $"Error in {GetType().Name}.HandleException: " + e.Message);
+                Logger.LogError(e, "Error in {name}.HandleException: {message}", GetType().Name, e.Message);
             }
         }
 
         public virtual void HandleException(ConsumeResult<K, V> consumeResult, Exception ex, string facilityId)
         {
-            var dlEx = new DeadLetterException(ex.Message, ex.InnerException);
+            var dlEx = new DeadLetterException(ex.Message, ex.InnerException ?? ex.GetBaseException());
```
smailliwcs: Maybe not as problematic as the above, but don't we want *this* exception (i.e., `ex` itself) to be the `DeadLetterException`'s inner exception?

DotNet/Report/Listeners/ResourceEvaluatedListener.cs
amphillipsLGC: Yeah agreed, that is why I added the additional log message. I think a bigger fix might be needed which would involve altering the Transient and Dead Letter exceptions in general which might be a bigger change than originally planned. I could do that in this scope of work though.

smailliwcs (approved):

PR #1033, opened 2025-08-01
TECH_DEBT: Adding requirements for KAM

seanmcilvenna:
### 🛠️ Description of Changes

Documentation updates

### 🧪 Testing Performed

N/A

### 🧑‍🔬 Unit Testing

N/A

### 📓 Documentation Updated

Added requirements to Knowledge Artifact Management domain.

## Summary by CodeRabbit

* **Documentation**
  * Reflowed front-matter formatting and expanded the Knowledge Artifact Management overview to enumerate artifact types.
  * Added a new "Requirements" section with Roles (including Quality Assurance Tester) and detailed Stakeholder & System requirements covering persistence, dependencies, multi-version handling, artifact registration and retrieval, terminology loading, and historical processing guarantees.
  * Added a Reporting requirement to retain IG version history for regenerating deleted reports.


docs/domains/KnowledgeArtifactManagement/index.mdx
```diff
@@ -23,4 +20,28 @@ This domain includes key artifacts such as:
...
+| As a quality analyst, I want to ensure that knowledge artifacts are correctly loaded and available to the Measure Evaluation, Validation, and Terminology services, so that measure evaluation processes are accurate and reliable. | Process                |
+| As a system administrator, I want the system to retain version history of IGs used during reporting, so that deleted reports can be accurately regenerated using their original definitions. | Report Reproducibility |
+
+### System Requirements
+
+| Requirement                                                                                                                                               | Category               |
+|-----------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------|
+| As the system, I want to register FHIR IG artifacts in a shared repository, so that multiple dependent services can access them without reloading.        | Process                |
```
xordroyd: For the "so that", I think reloading is outside the scope of the KAM domain, and that those services dependent on KAM (e.g., Evaluation, Compliance) should take on that responsibility. 

Consider "so that dependent system functionality can access and use appropriately" or something like that?

docs/domains/KnowledgeArtifactManagement/index.mdx
```diff
@@ -23,4 +20,28 @@ This domain includes key artifacts such as:
...
+| As the system, I want to retrieve version-specific Measure artifacts from the knowledge base, so that I evaluate patients against the correct logic.      | Version Management     |
```
xordroyd: I like "access" better than "retrieve". 

For the "so that", knowledge artifacts can extend beyond evaluation of correct logic. Maybe it's a general statement like "so that appropriate reporting and other requirements can be established and assured."

docs/domains/KnowledgeArtifactManagement/index.mdx
```diff
@@ -23,4 +20,28 @@ This domain includes key artifacts such as:
...
+| As the system, I want to access profile definitions from the loaded IGs, so that I can validate resource instances accurately.                            | Process                |
```
xordroyd: Is it fine to _not_ specify FHIR when we use FHIR nouns (Profile, IG, Resource)? I think it is, but would prefer to capitalize them to convey that they are FHIR terms.

docs/domains/KnowledgeArtifactManagement/index.mdx
```diff
@@ -23,4 +20,28 @@ This domain includes key artifacts such as:
...
+| As the system, I want to load and reference ValueSet and CodeSystem definitions from IGs, so that I can confirm inform other services on the terminology needed to accurately evaluate and validate.          | Process                |
+| As the system, I want to store IG version metadata and associate it with report evaluation records, so that report reconstruction uses the correct logic. | Report Reproducibility |
```
xordroyd: Should we say "inject" instead of "store" to connote it being submission metadata?  Also, this would be more general than just evaluation. Would "reporting processes" work to generalize? 

For "so that", the version could be used beyond reconstruction logic, so a more general statement may be appropriate. "So that report processes may be examinable, verifiable, and reproduceable."

docs/domains/KnowledgeArtifactManagement/index.mdx
```diff
@@ -23,4 +20,28 @@ This domain includes key artifacts such as:
...
+| As the system, I want to retrieve historical IG versions from a versioned repository, so that report regeneration uses the same logic and constraints.    | Report Reproducibility |
+| As the system, I want to enable regeneration of deleted reports using audit logs and archived IGs, so that no reportable event is permanently lost.       | Report Reproducibility |
+| As the system, I want to ensure IG updates do not overwrite previously loaded versions, so that all versions remain available for historical processing.  | Version Management     |
```
xordroyd: This requirements seems superfluous to the req on line 44.

docs/domains/KnowledgeArtifactManagement/index.mdx
```diff
@@ -23,4 +20,28 @@ This domain includes key artifacts such as:
...
\ No newline at end of file
+Rather than owning executable services, this domain captures the processes, standards alignment, and content lifecycle required to ensure consistency and reusability across the system.
+
+## Requirements
+
+### Stakeholder Requirements
+
+| Requirement                                                                                                                                               | Category               |
+|-----------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------|
+| As a system administrator, I want to load FHIR Implementation Guides (IGs) that include Measures, Profiles, and Terminology artifacts, so that the platform can support standards-based evaluation and validation. | Process                |
```
xordroyd: This is the heart of the whole system in my mind. Will there always be an IG? I kinda want to say yes, though we've contemplated scenarios where we might not explicitly require one. 
Thoughts: 

1. "I want to _load_ FHIR IG into the KA repository" ... is load the right verb here? I think so... but also have been thinking about "specify" or "store". 
2. I think while the IG is primary importance to the system as a whole, the KAM domain's main purpose is access management of specified KAs, so is the _first_ user story something like "As a knowledge artifact administrator (specialization of systems admin), I can manage persistence, access, relationship and other characteristics of knowledge artifacts so that dependent functionality operates appropriately." 
3. Another key concept here, that I am unsure how to weave in, is the idea of "modularizing" or "SIFI" (separating interface from impl) of standards/KAs. By having this system we achieve the ability to reference something that can change--like Eval is always following "latest" but that collapses to "v20250811" when there's an eval in a couple weeks. 
4. Are we good with the term Measure? dQMs and CQL, Measure and Metric all are thrown around with different meaning. https://build.fhir.org/clinicalreasoning-knowledge-artifact-representation.html

docs/domains/KnowledgeArtifactManagement/index.mdx
```diff
@@ -23,4 +20,28 @@ This domain includes key artifacts such as:
...
+| As a developer, I want to manage multiple versions of the same IG, so that I can support version-specific behavior and legacy compatibility.              | Version Management     |
```
xordroyd: All knowledge artifacts, and (maybe) the dependency graph between them. Also, the "so that" is not general enough.

docs/domains/KnowledgeArtifactManagement/index.mdx
```diff
@@ -23,4 +20,28 @@ This domain includes key artifacts such as:
...
+| As a quality analyst, I want to ensure that knowledge artifacts are correctly loaded and available to the Measure Evaluation, Validation, and Terminology services, so that measure evaluation processes are accurate and reliable. | Process                |
```
xordroyd: We need a ubiquitouslanguage doc that defines these roles. What's a "quality analyst"? 

Beyond that, "loaded correctly" isn't useful. We don't need requirements that check correctness of everything. Better to specify the positive: "As a quality analyst, I can specify knowledge artifact configuration so that I satisfy reporting requirements."

If this is a compliance user story, that should be under the compliance domain.

docs/domains/KnowledgeArtifactManagement/index.mdx
```diff
@@ -23,4 +20,28 @@ This domain includes key artifacts such as:
...
+| As a system administrator, I want the system to retain version history of IGs used during reporting, so that deleted reports can be accurately regenerated using their original definitions. | Report Reproducibility |
```
xordroyd: System admin? I think this is a "as a Reporting Organization and as a Report Consumer" 

Also, this seems it should be under the Report domain, not here.

docs/domains/KnowledgeArtifactManagement/index.mdx
seanmcilvenna: #1 I like "load" in this case, myself, but good thought. Added #2 to the list of requirements as I think it's a good one. #3 I like what you're describing but can't think of how to represent it here. So, I'm going to leave it for now. #4 Put some thought into it... I prefer Measure here, as it is actually a FHIR resource type, which makes sense for this particular sentence.

docs/domains/KnowledgeArtifactManagement/index.mdx
seanmcilvenna: Added a requirement to address this.

docs/domains/KnowledgeArtifactManagement/index.mdx
seanmcilvenna: Didn't like the entire story, ended up removing it.

docs/domains/KnowledgeArtifactManagement/index.mdx
seanmcilvenna: I prefer "store" for now, because they aren't going to use that injected version information when they ingest the submission; at least not right now. Later, we may want to add one for "inject". I just want this to convey that we're at least capturing/persisting the necessary information to support "injecting" later (if we choose).

xordroyd (approved):
This is a great functional baseline for our KAM capability!!

docs/domains/Report/index.mdx
```diff
@@ -38,6 +38,7 @@ reports. Each service plays a specialized role in enabling modularity, traceabil
 * Reporting functionality SHALL be modular so that it can be used for multiple use cases
 * Link acquires (either via PUSH or PULL) and evaluates the data in a reporting pipeline.
 * Reporting pipelines SHALL be orchestrated via configuration of services
+* As a reporting organization and reporting consumer, I want the system to retain version history of IGs used during reporting in a reporting period, so that deleted reports can be accurately regenerated using their original definitions.
```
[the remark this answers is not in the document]
seanmcilvenna: Ignoring nitpick

docs/domains/KnowledgeArtifactManagement/index.mdx
```diff
@@ -23,4 +20,34 @@ This domain includes key artifacts such as:
...
+| As the system, I want to retrieve version-specific Measure artifacts from the knowledge artifact repository, so that reporting requirements can be confirmed.                                        | Version Management     |
+| As the system, I want to access profile definitions from the loaded IGs, so that I can validate resource instances accurately.                                                                       | Process                |
+| As the system, I want to load and reference ValueSet and CodeSystem definitions from IGs, so that I can confirm inform other services on the terminology needed to accurately evaluate and validate. | Process                |
+| As the system, I want to store IG version metadata and associate it with report evaluation records, so that report reconstruction uses the correct logic.                                            | Report Reproducibility |
+| As the system, I want to ensure IG updates do not overwrite previously loaded versions, so that all versions remain available for historical processing.                                             | Version Management     |
```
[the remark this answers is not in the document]
seanmcilvenna: Ignoring nitpick

PR #1116, opened 2025-09-18
TECH_DEBT: Pipeline Update

kissalk:
Pipeline Update


michael-misiaszek (changes requested):
This looks like it's good too. However, we need to ensure we package up all changed needed for this pipeline. Are you still needing to add the changes for the appsettings.json? These settings would be to change the "Verstion" & "Git-Commit" sections in that file.

PR #1237, opened 2025-11-05
LNK-4316: Remove Namespace and Project as $type data from QueryPlan API

nvmLantana:
This update removes the requirement to send API model namspaces and project names as $type data in the API call. A Caller should not have to know our internal namespace structure when calling our API's. Futhermore, changing namespaces will require the API users to be aware of those name space changes, or else the API will "break", which happened as a result of the POI shuffling of models into new directories.

Old:
image

New:
image

ParameterType Options:
image

QueryConfigType Options:
image

### 🛠️ Description of Changes
Please provide a high-level overview of the changes included in this PR.

### 🧪 Testing Performed

### 🧑‍🔬 Unit Testing

### 📓 Documentation Updated


arianamihailescu (approved):
Looks good

PR #1239, opened 2025-11-06
LNK-4303: add event date

edward-miller-lcg:
### 🛠️ Description of Changes
Add EventDate column to PatientEvents to simplify querying when the need to evaluate an admit or discharge date. Update integration test for bad assertion. Update migration to backfill EventDates based on admit or discharge dates.

### 🧪 Testing Performed
Local testing and integration testing.

### 🧑‍🔬 Unit Testing

### 📓 Documentation Updated
n/a

## Summary by CodeRabbit

* **New Features**
  * Patient events now capture and track actual event dates based on admission or discharge timing.

* **Improvements**
  * Updated event queries to sort and filter by actual event dates rather than creation timestamps, ensuring accurate chronological ordering and reporting of patient encounters.

* **Tests**
  * Test data adjusted to validate updated event date handling.


DotNet/Census/Migrations/20251106152644_AddEventDateToPatientEvent.cs
```diff
@@ -0,0 +1,72 @@
...
+                        CreateDate  -- fallback to CreateDate if no specific date in payload
+                END
+            WHERE EventType IN ('FHIRListAdmit', 'FHIRListDischarge')
+               OR EventType IS NULL");
+
+            // Handle any remaining NULLs (optional: set to CreateDate)
+            migrationBuilder.Sql(@"
+            UPDATE PatientEvents
+            SET EventDate = CreateDate
+            WHERE EventDate IS NULL");
```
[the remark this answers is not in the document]
edward-miller-lcg: I tested the backfill against system generated rows and it ran just fine.

DotNet/Census/Application/Factories/PatientEventFactory.cs
```diff
@@ -18,7 +19,18 @@ public static PatientEvent Create(string correlationId, string patientId, string
...
+        };
+    }
+
+    private static DateTime GetEventDate(IPayload payload)
+    {
+        return payload switch
+        {
+            FHIRListDischargePayload dischargePayload => dischargePayload.DischargeDate,
+            FHIRListAdmitPayload admitPayload => admitPayload.AdmitDate,
+            _ => DateTime.UtcNow
```
smailliwcs: Is this a legitimate situation?  I.e., do we have other `IPayload`s aside from the above two?  If not, should we `throw` here instead?  That way, if we *do* add another `IPayload` in the future, we get an error instead of a value that's probably incorrect.

smailliwcs (approved):

DotNet/Census/Application/Factories/PatientEventFactory.cs
edward-miller-lcg: I agree that throwing makes the most sense here. Especially as we add to this for different source types and how that EventDate can be radically different from DateTime.UtcNow.

edward-miller-lcg: change made.

PR #1243, opened 2025-11-06
LNK-4366: Submission Service: Pass Query Parameter to Endpoint, Default to Internal ABS

smailliwcs:
### 🛠️ Description of Changes
Support download from both (internal and external) blob containers through the Submission API.  By default, download from internal (since submissions will be removed from external after downstream ingestion).  But for smoke testing, download from external in order to test full end-to-end flow.

### 🧪 Testing Performed
Tested locally (and as part of automated smoke testing).

### 🧑‍🔬 Unit Testing
N/A

### 📓 Documentation Updated
N/A

## Summary by CodeRabbit

* **New Features**
  * Download endpoints now support an optional parameter to specify storage source (internal or external), enabling flexible report retrieval from either location.


DotNet/Submission/Controllers/SubmissionController.cs
```diff
@@ -30,10 +30,14 @@ public class SubmissionController(
      * Downloads the specified report's data as a ZIP archive
      * <param name="facilityId">The ID of the facility</param>
      * <param name="reportId">The ID of the report to download</param>
+     * <param name="external">Whether to download from external or internal (default) storage</param>
      * <remarks>Gets information about the report from the report service in order to construct the directory/path for the report.</remarks>
      */
     [HttpGet("{facilityId}/{reportId}")]
-    public async Task<IActionResult> DownloadReport([FromRoute] string facilityId, [FromRoute] string reportId)
+    public async Task<IActionResult> DownloadReport(
+        [FromRoute] string facilityId,
```
edward-miller-lcg, line 38: These attributes typically get flagged by static scanning for binding. I think the recommendation is to not explicitly tag the route parameters.

edward-miller-lcg (approved):

DotNet/Submission/Controllers/SubmissionController.cs
smailliwcs: Good point.  Confirmed with an ad hoc scan that this PR results in no new issues.  Looking back at some older scans, it seems like "Mass Assignment: Insecure Binder Configuration" only kicks in for unconditional binding to a complex object.

PR #1272, opened 2025-11-19
release/0.3.0-bundling-efficiency-patch

seanmcilvenna:
### 🛠️ Description of Changes

Changing strategy to using batch method for both shared and patient resources.
Wrapping the batched retrieval of resources in threads so that each batch is not running synchronously; turning ~3 seconds into ~0.5 seconds.

### 🧪 Testing Performed

Ran locally end-to-end with a patient that has over 10k resources, saw the patient successfully submitted.

### 🧑‍🔬 Unit Testing

### 📓 Documentation Updated

Updated documentation to include new config property. Config property has default values and does not need to be specified at deployment time unless the default needs to be overridden.

## Summary by CodeRabbit

* **New Features**
  * Added configurable option to control resource collection concurrency settings.

* **Chores**
  * Updated infrastructure configuration to increase message size limits for Kafka.
  * Removed debug-level logging from default configuration.
  * Updated documentation to reflect new configuration options and settings.

### DIFF
--- a/Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/repositories/AbstractResourceRepository.java
+++ b/Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/repositories/AbstractResourceRepository.java
@@ -1,11 +1,15 @@
 package com.lantanagroup.link.measureeval.repositories;
 
+import com.lantanagroup.link.measureeval.configs.LinkConfig;
 import com.lantanagroup.link.measureeval.entities.AbstractResourceEntity;
 import com.lantanagroup.link.measureeval.entities.PatientReportingEvaluationStatus;
 import com.lantanagroup.link.measureeval.entities.PatientResource;
 import com.lantanagroup.link.measureeval.entities.SharedResource;
 import com.lantanagroup.link.measureeval.records.AbstractResourceRecord;
+import com.lantanagroup.link.measureeval.services.PatientStatusBundler;
 import org.hl7.fhir.r4.model.ResourceType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.data.mongodb.core.MongoOperations;
 import org.springframework.data.mongodb.core.query.Criteria;
 import org.springframework.data.mongodb.core.query.Query;
@@ -13,15 +17,21 @@ import org.springframework.stereotype.Repository;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.*;
+import java.util.stream.Collectors;
 
 import static org.springframework.data.mongodb.core.query.Criteria.byExample;
 
 @Repository
 public class AbstractResourceRepository {
-    private final MongoOperations mongoOperations;
+    private static final Logger logger = LoggerFactory.getLogger(AbstractResourceRepository.class);
 
-    public AbstractResourceRepository(MongoOperations mongoOperations) {
+    private final MongoOperations mongoOperations;
+    private final LinkConfig linkConfig;
+
+    public AbstractResourceRepository(MongoOperations mongoOperations, LinkConfig linkConfig) {
         this.mongoOperations = mongoOperations;
+        this.linkConfig = linkConfig;
     }
 
     public AbstractResourceEntity findOne(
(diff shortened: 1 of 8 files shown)


Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/repositories/AbstractResourceRepository.java
```diff
@@ -63,53 +68,68 @@ public <T extends AbstractResourceEntity> T save(T entity) {
...
-                        .and("resourceType").is(resource.getResourceType())
-                        .and("resourceId").is(resource.getResourceId()))
-                .toList();
-
-        Criteria combinedCriteria = new Criteria().orOperator(criteriaList.toArray(new Criteria[0]));
-        Query query = new Query(combinedCriteria);
+        List<CompletableFuture<List<AbstractResourceEntity>>> futures = new ArrayList<>();
+        boolean useConfiguredThreadCount = linkConfig.getMaxCollectResourcesThreads() != null &&
+                linkConfig.getMaxCollectResourcesThreads() > 0 &&
+                linkConfig.getMaxCollectResourcesThreads() < Runtime.getRuntime().availableProcessors();
```
smailliwcs, line 79: Since the work we're doing here not CPU-bound, it _could_ make sense to have the number of threads exceed the number of available processors.  Regardless, if we want to impose that limit, why not _clamp_ the number of threads to it, rather than discarding an out-of-range value?

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/repositories/AbstractResourceRepository.java
```diff
@@ -63,53 +68,68 @@ public <T extends AbstractResourceEntity> T save(T entity) {
...
+        int threadCount = Runtime.getRuntime().availableProcessors();
 
-        return mongoOperations.find(query, entityType);
-    }
+        if (useConfiguredThreadCount) {
+            logger.debug("Using configured thread count of {}", linkConfig.getMaxCollectResourcesThreads());
+        } else {
```
smailliwcs, line 85: As mentioned below, we should be assigning the configured thread count here.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/repositories/AbstractResourceRepository.java
```diff
@@ -63,53 +68,68 @@ public <T extends AbstractResourceEntity> T save(T entity) {
...
+            List<PatientReportingEvaluationStatus.Resource> remainingResources = new ArrayList<>(resourceReferences);
+            while (!remainingResources.isEmpty()) {
+                int characterCount = 0;
+                List<Criteria> batchCriteria = new ArrayList<>();
+
+                for (int i = remainingResources.size() - 1; i >= 0; i--) {
+                    PatientReportingEvaluationStatus.Resource resource = remainingResources.get(i);
+                    int resourceCharLength = resource.getResourceType().toString().length() + resource.getResourceId().length();
+
+                    if (characterCount + resourceCharLength >= 10000) {
```
smailliwcs, line 102: Do we need the character count limit to be configurable?  Maybe not, if 10K characters is well below any kind of default limit we'd see across environments.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/repositories/AbstractResourceRepository.java
```diff
@@ -63,53 +68,68 @@ public <T extends AbstractResourceEntity> T save(T entity) {
...
+                    characterCount += resourceCharLength;
+                    batchCriteria.add(Criteria.where("facilityId").is(facilityId)
+                            .and("resourceType").is(resource.getResourceType())
+                            .and("resourceId").is(resource.getResourceId()));
+                    remainingResources.remove(i);
+                }
+
+                if (!batchCriteria.isEmpty()) {
+                    List<Criteria> finalBatchCriteria = batchCriteria;
+                    CompletableFuture<List<AbstractResourceEntity>> future = CompletableFuture.supplyAsync(() -> {
```
smailliwcs, line 115: Could consider `executor.submit` here, though that returns a plain `Future`, which would complicate how we aggregate the results below.  Specifically, we'd need to handle `InterruptedException` and `ExecutionException`.  Which we _would_ want to do if this were a long-term solution, but maybe not worth the effort for a stopgap fix.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/repositories/AbstractResourceRepository.java
```diff
@@ -63,53 +68,68 @@ public <T extends AbstractResourceEntity> T save(T entity) {
...
+                        Criteria combinedCriteria = new Criteria().orOperator(finalBatchCriteria.toArray(new Criteria[0]));
+                        Query query = new Query(combinedCriteria);
+                        List<? extends AbstractResourceEntity> results = isShared ?
+                                mongoOperations.find(query, SharedResource.class) :
+                                mongoOperations.find(query, PatientResource.class);
+                        return new ArrayList<>(results);
```
smailliwcs, line 121: Why create the  `ArrayList` here?  I.e., can we return `results` as is?  Maybe be changing `futures` to a `List<CompletableFuture<List<? extends AbstractResourceEntity>>>`?

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/repositories/AbstractResourceRepository.java
```diff
@@ -63,53 +68,68 @@ public <T extends AbstractResourceEntity> T save(T entity) {
...
+                }
+            }
+
+            return futures.stream()
+                    .map(CompletableFuture::join)
+                    .flatMap(List::stream)
+                    .collect(Collectors.toList());
+        } finally {
+            executor.shutdown();
+        }
```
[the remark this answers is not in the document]
smailliwcs: Disagree with the CodeRabbit comment here ... `shutdown` just _initiates_ a shutdown in a non-blocking way (_without_ waiting for tasks to complete).  Though that introduces the "opposite" risk ... namely, that we could conceivably exit this method while tasks are still running (if one of them threw an exception, causing `join` to fail).  Again, maybe not worth handling given that this is a short-term solution.

Java/measureeval/src/main/java/com/lantanagroup/link/measureeval/services/AbstractResourceConsumer.java
```diff
@@ -115,7 +115,14 @@ protected void doConsume (String correlationId, ConsumerRecord<String, T> record
             if (logger.isInfoEnabled()) {
                 logger.info("Consuming record: RECORD=[{}] FACILITY=[{}] CORRELATION=[{}] ACQUISITION COMPLETE=[{}]", KafkaUtils.format(record), facilityId, correlationId, value.isAcquisitionComplete());
             }
-            PatientReportingEvaluationStatus patientStatus = Objects.requireNonNullElseGet(retrievePatientStatus(facilityId, correlationId), () -> createPatientStatus(facilityId, correlationId, value));
+
+            PatientReportingEvaluationStatus patientStatus = retrievePatientStatus(facilityId, correlationId);
+
+            if (patientStatus == null) {
+                logger.warn("Patient status for facilityId: {}, correlationId: {} not found. Creating a temporary PatientStatus...", facilityId, correlationId);
```
smailliwcs, line 122: A null `patientStatus` isn't really an exceptional condition ... it just means we're ingesting the first message for a particular correlation ID.  (And `createPatientStatus` takes care of creating it in persistence.)

smailliwcs (changes requested):

PR #1378, opened 2026-01-16
LNK-4430: Data Acquisition - Handle 429 Status

nvmLantana:
### 🛠️ Description of Changes
The changes implement robust handling for HTTP 429 (Too Many Requests) responses in the Data Acquisition service, adhering to RFC 6585 by respecting the Retry-After header or falling back to exponential backoff. This includes detecting 429 in FHIR API calls, propagating a custom exception, and applying retry logic with delays in log processing. Additionally, prioritization ensures ACH-Daily queries (set to High priority) process before ACH-Monthly (Normal priority) via updated query ordering. Retries are capped at 5 (increased from 1), and logs are rescheduled by resetting ExecutionDate and Status to Pending. A shared utility class handles header capture for consistency. These enhancements prevent server overload, prioritize critical daily reports, and minimize data loss from rate limiting, with minimal impact on existing code.
Per Class/File Summary

**TooManyRequestsException.cs** (New File):
Added a custom exception inheriting from FhirOperationException to encapsulate 429 errors, including an optional RetryAfter TimeSpan property for delay information.

**FhirCommandUtils.cs** (New File):
Introduced a shared utility class containing the HeaderCapturingHandler (a DelegatingHandler to capture HTTP response headers) and the static ParseRetryAfter method to extract and parse the Retry-After header into a TimeSpan.

**ReadFhirCommand.cs** (Modified):
Integrated HeaderCapturingHandler to wrap the HttpClient for header capture. In ExecuteAsync, catch FhirOperationException for 429, parse Retry-After using the utility, and throw TooManyRequestsException. Used a new HttpClientHandler as the inner handler for the chain.

**SearchFhirCommand.cs** (Modified):
Similar to ReadFhirCommand.cs: Added HeaderCapturingHandler integration in ExecuteAsync and ExecuteNonPagingAsync. Catch 429 in search and paging operations, parse Retry-After, and throw TooManyRequestsException. Applied to initial search, paging continuation, and non-paging methods.

**PatientDataService.cs** (Modified):
In ExecuteLogRequest, added catch for TooManyRequestsException: Increment RetryAttempts, calculate delay (using RetryAfter or exponential backoff capped at 60s), reset ExecutionDate and Status to Pending, update log notes, and return to requeue. If retries exceed max, set Failed. In CreateLogEntries, set Priority to High for "Daily" reports, Normal otherwise.

**DataAcquisitionLog.cs** (Modified):
Increased MaxRetryAttempts constant from 1 to 5 to allow more retries before permanent failure.

**FhirApiService.cs** (Modified):
Added catches in ExecuteRead and ExecutePagingSearch to propagate TooManyRequestsException while handling other FhirOperationException as before (e.g., ignoring NotFound for references).

**DataAcquisitionLogQueries.cs** (Modified):
Updated GetNextEligibleBatchForFacility query. Added a filter to only include logs who either have a null execution date, or their execution date is present or past.

### 🧪 Testing Performed

### 🧑‍🔬 Unit Testing

### 📓 Documentation Updated

## Summary by CodeRabbit

* **New Features**
  * Added rate-limiting (429) response handling with automatic retry-after support
  * Introduced intelligent retry mechanism with exponential backoff (up to 60 seconds)
  * Implemented dynamic priority assignment for daily reports

* **Improvements**
  * Increased max retry attempts from 1 to 5 for better resilience
  * Enhanced batch eligibility logic to include retry-limited failed items
  * Added retry delay calculation based on attempt count or server-provided retry-after headers


DotNet/DataAcquisition.Domain/Application/Services/PatientDataService.cs
```diff
@@ -533,12 +538,39 @@ await _dataAcquisitionLogManager.UpdateAsync(new UpdateDataAcquisitionLogModel
...
+            catch (TooManyRequestsException ex)
+            {
+                _logger.LogWarning(ex, "Throttled by 429 for facility {FacilityId}", log.FacilityId.Sanitize());
+
+                log.Notes ??= new List<string>();
+
+                log.RetryAttempts ??= 0;
+
+                log.ExecutionDate = DateTime.UtcNow.Add(ex.RetryAfter);
+                log.Status = RequestStatus.Pending; //Don't count this as a failure
```
smailliwcs, line 583: Why not count as a failure?  If the server keeps sending HTTP 429s, the report will never complete.  And if the server is legitimately overloaded through `MaxRetryAttempts`, we probably don't want to keep hammering on it.

smailliwcs (commented):
IIUC, this makes it so that, if we receive an HTTP 429, we defer the _current_ DA log in a way that respects the `Retry-After` header.  Don't we need something that defers _all_ DA logs (for the current tenant)?

DotNet/DataAcquisition.Domain/Application/Services/PatientDataService.cs
nvmLantana: If we have a lot of data to get through their system, and we are overloading it, counting as failures will lead to potentially massive holes in the data. Presumably, we do want to get the data, at whatever pace they designate is appropriate for us to do so, and keep trying until we get it. So, we either try and get the data we are supposed to get, eventually, or we quietly produce a busted report. I assumed that if we are getting a 429, they want us to retry later...

smailliwcs (approved):

PR #1379, opened 2026-01-16
LNK-4714: Replicate MVP Report Output

edward-miller-lcg:
### 🛠️ Description of Changes
Replcates MVP output in BOTW based on flag useNdJson (bool).

### 🧪 Testing Performed
Tested at connectathon

### 🧑‍🔬 Unit Testing

### 📓 Documentation Updated

## Release Notes

* **New Features**
  * Added configurable workflow toggle to switch between NDJSON and expanded bundle processing strategies for data submissions.
  * Enhanced JSON parser to support more lenient FHIR resource validation.

* **Chores**
  * Removed formatting artifact from source file.


DotNet/Submission/Application/Services/BlobStorageService.cs
```diff
@@ -112,24 +147,542 @@ public async Task UploadToExternalAsync(
...
+                string measureName = fallbackAgg.Measure.Split('/').Last().Split('|')[0];
+                string aggregateName = $"aggregate-{measureName}.json";
+                string blobName = GetBlobName(_externalSettings.BlobRoot, measureFolder, reportName, aggregateName);
+                _logger.LogDebug("Fallback uploading aggregate: {BlobName}", blobName);
+
+                BlockBlobClient blobClient = _externalContainerClient!.GetBlockBlobClient(blobName);
+                BlockBlobOpenWriteOptions blobOptions = new() { HttpHeaders = new BlobHttpHeaders { ContentType = "application/json" } };
+                await using var stream = await blobClient.OpenWriteAsync(true, blobOptions, cancellationToken);
+                await stream.WriteAsync(Encoding.UTF8.GetBytes(JsonSerializer.Serialize(fallbackAgg, jsonOptions)), cancellationToken);
+            }
```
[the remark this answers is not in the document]
edward-miller-lcg: good catch. fixed.

DotNet/Normalization/Application/FhirJsonConverter.cs
```diff
@@ -10,6 +10,7 @@ public class FhirResourceConverter : JsonConverter<DomainResource>
 
     public override DomainResource Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
     {
+        _fhirParser.Settings.PermissiveParsing = true;
```
smailliwcs, line 13: Is this related to/required for JSON output?

DotNet/Submission/Application/Services/BlobStorageService.cs
smailliwcs, line 1: Would it be possible to move the JSON translation logic into a service that `BlobStorageService` can take as a dependency?

DotNet/Submission/Application/Services/BlobStorageService.cs
```diff
@@ -39,13 +46,15 @@ private static string GetBlobName(string? blobRoot, params string[] segments)
...
-            IOptions<ExternalBlobStorageSettings> externalSettings)
+            IOptions<ExternalBlobStorageSettings> externalSettings,
+            IConfiguration configuration)
         {
             _logger = logger;
             _internalSettings = internalSettings.Value;
             _externalSettings = externalSettings.Value;
             _internalContainerClient = GetContainerClient(_internalSettings);
             _externalContainerClient = GetContainerClient(_externalSettings);
+            _useNdJson = configuration.GetValue<bool>("useNdJson");
```
smailliwcs, line 57: Recommend moving this out of top-level configuration.  Maybe under `ExternalBlobStorage`?  Or else a new section dedicated to the JSON translation service (if you end up creating that).

DotNet/Submission/Application/Services/BlobStorageService.cs
```diff
@@ -112,24 +145,548 @@ public async Task UploadToExternalAsync(
...
+            SubmitPayloadKey key,
+            SubmitPayloadValue value,
+            byte[] content,
+            CancellationToken cancellationToken = default)
+        {
+            _logger.LogInformation("Processing manifest for expanded bundles - ReportScheduleId: {ReportScheduleId}", key.ReportScheduleId);
+
+            // Parse manifest NDJSON
+            string manifestContent = Encoding.UTF8.GetString(content);
+            var jsonOptions = new JsonSerializerOptions().ForFhir(new FhirJsonPocoDeserializerSettings { Validator = null });
```
smailliwcs, line 173: Firely recommends caching instances of `JsonSerializerOptions` statically for performance.

DotNet/Submission/Application/Services/BlobStorageService.cs
```diff
@@ -112,24 +145,548 @@ public async Task UploadToExternalAsync(
...
+            else
+            {
+                _logger.LogWarning("No aggregate reports detected in manifest after parsing.");
+            }
+
+            string measureFolder = GetMeasureFolderPath(value.ReportTypes);
+            string reportName = ReportHelpers.GetReportName(key.ReportScheduleId, key.FacilityId, value.ReportTypes, value.StartDate);
+
+            string? nhsnOrgId = organization?.Identifier
+                .FirstOrDefault(i => i.System == "https://www.cdc.gov/nhsn/OrgID")?.Value;
```
smailliwcs, line 223: The NHSN OrgID identifier system is already a constant in Report's `ReportConstants.BundleSettings`.  Consolidate those with anything that should be a constant _here_, then move everything to Shared?

DotNet/Submission/Application/Services/BlobStorageService.cs
```diff
@@ -112,24 +145,548 @@ public async Task UploadToExternalAsync(
...
+            }
+            else
+            {
+                _logger.LogWarning("No OperationOutcome found in manifest - skipping validation.json upload");
+            }
+
+            _logger.LogInformation("Completed processing and uploading {PatientCount} expanded patient bundles", patientIds.Count);
+        }
+
+        private static string PatchEmptyDeviceVersionValue(string json)
```
smailliwcs, line 529: Do we need this?  If so, can we do it in `ReportManifestProducer.CreateDevice`?  (And can we use Firely instead of direct JSON manipulation?)

DotNet/Submission/Application/Services/BlobStorageService.cs
seanmcilvenna, line 57: Use `UseNdJson` casing instead of `useNdJson`

seanmcilvenna (changes requested):

DotNet/Normalization/Application/FhirJsonConverter.cs
edward-miller-lcg: Permissive Parsing is to allow for invalid json objects to make it through as a C# object. Otherwise, if it's invalid FHIR, it will throw an error.

smailliwcs: Guess my question was more ... does MVP-style report output require Normalization to parse more leniently?  Seems unrelated to the rest of the PR.

DotNet/Submission/Application/Services/BlobStorageService.cs
smailliwcs: Side note: there's now a lenient `JsonSerializerOptions` instance available for project-wide use in Shared's `LinkFhirSerializerOptions`.

DotNet/Submission/Application/Services/BlobStorageService.cs
```diff
@@ -112,24 +145,548 @@ public async Task UploadToExternalAsync(
...
+                    if (!string.IsNullOrEmpty(refId))
+                    {
+                        patientIds.Add(refId);
+                    }
+                }
+            }
+            else
+            {
+                _logger.LogWarning("No patient List found in manifest. Falling back to file names.");
+            }
```
smailliwcs, line 265: Do we need to consult the census?  We can only submit patients that we have files for, not to mention that we'll _expect_ missing files in some cases (specifically, for non-reportable patients).

DotNet/Submission/Application/Services/BlobStorageService.cs
```diff
@@ -168,5 +725,33 @@ public Task<IDictionary<string, byte[]>> DownloadFromExternalAsync(string payloa
             string prefix = ChangeBlobRoot(uriBuilder.BlobName);
             return DownloadAsync(_externalContainerClient, prefix, cancellationToken);
         }
+
+        private string GetMeasureAcronym(List<string> reportTypes)
```
smailliwcs, line 729: Use Shared's `MeasureNameShortener` instead?

DotNet/Submission/Application/Services/BlobStorageService.cs
```diff
@@ -112,24 +145,548 @@ public async Task UploadToExternalAsync(
...
+                    {
+                        _logger.LogWarning(ex, "Failed to parse line in patient {PatientId} NDJSON", patientId);
+                    }
+                }
+
+                _logger.LogDebug("Parsed {Count} resources for patient {PatientId}", patientResources.Count, patientId);
+
+                // Find the individual MeasureReport in patient resources
+                var patientMR = patientResources.OfType<MeasureReport>()
+                    .FirstOrDefault(mr => !IsAggregateMeasureReport(mr) && mr.Subject?.Reference?.Contains(patientId) == true);
```
smailliwcs, line 311: This doesn't seem to handle the multi-measure use case (where there could be multiple individual measure reports in a patient file).

DotNet/Submission/Application/Services/BlobStorageService.cs
```diff
@@ -112,24 +145,548 @@ public async Task UploadToExternalAsync(
...
+                // Create expanded bundle with patient resources only (no shared resources from manifest)
+                var expandedBundle = new Bundle
+                {
+                    Type = Bundle.BundleType.Collection,
+                    Timestamp = DateTimeOffset.UtcNow
+                };
+
+                expandedBundle.Entry.Add(new Bundle.EntryComponent
+                {
+                    FullUrl = $"http://www.cdc.gov/nhsn/fhirportal/dqm/ig/MeasureReport/{patientMR.Id}",
```
smailliwcs, line 330: Consider defining a constant for the URL prefix.

DotNet/Submission/Application/Services/BlobStorageService.cs
```diff
@@ -112,24 +145,548 @@ public async Task UploadToExternalAsync(
...
+
+                // Add Organization to each patient bundle
+                if (organization != null)
+                {
+                    expandedBundle.Entry.Add(new Bundle.EntryComponent
+                    {
+                        FullUrl = $"http://www.cdc.gov/nhsn/fhirportal/dqm/ig/Organization/{organization.Id}",
+                        Resource = organization
+                    });
+                }
```
smailliwcs, line 342: IIRC, in MVP, we only put the `Organization` in submitting-org.json ... we _don't_ also copy into each patient file.

DotNet/Submission/Application/Services/BlobStorageService.cs
```diff
@@ -112,24 +145,548 @@ public async Task UploadToExternalAsync(
...
+                BlockBlobOpenWriteOptions blobOptions = new()
+                {
+                    HttpHeaders = new BlobHttpHeaders { ContentType = "application/json" }
+                };
+
+                await using var stream = await blobClient.OpenWriteAsync(true, blobOptions, cancellationToken);
+                await stream.WriteAsync(aggregateBytes, cancellationToken);
+            }
+
+            if (!aggregateReports.Any() && manifestResources.OfType<MeasureReport>().Any(m => m.Type?.ToString()?.ToLowerInvariant().Contains("subject-list") == true))
```
smailliwcs, line 392: Do we need this fallback?  Recommend simplifying `IsAggregateMeasureReport` to "Is `MeasureReport.type` equal to `subject-list`?" and removing this block.

DotNet/Submission/Application/Services/BlobStorageService.cs
```diff
@@ -112,24 +145,548 @@ public async Task UploadToExternalAsync(
...
+                await using var stream = await blobClient.OpenWriteAsync(true, blobOptions, cancellationToken);
+                await stream.WriteAsync(aggregateBytes, cancellationToken);
+            }
+
+            // shared-resources.json - now an EMPTY Bundle
+            var sharedBundle = new Bundle
+            {
+                Type = Bundle.BundleType.Collection,
+                Timestamp = DateTimeOffset.UtcNow
+            };
```
smailliwcs, line 415: Don't we need to add the shared resources here since we're not adding them to the patient bundles above?

DotNet/Submission/Application/Services/BlobStorageService.cs
```diff
@@ -112,24 +145,548 @@ public async Task UploadToExternalAsync(
...
+                _logger.LogDebug("Uploading expanded patient bundle: {BlobName}", blobName);
+
+                BlockBlobClient blobClient = _externalContainerClient!.GetBlockBlobClient(blobName);
+                BlockBlobOpenWriteOptions blobOptions = new()
+                {
+                    HttpHeaders = new BlobHttpHeaders { ContentType = "application/json" }
+                };
+
+                await using var stream = await blobClient.OpenWriteAsync(true, blobOptions, cancellationToken);
+                await stream.WriteAsync(bundleBytes, cancellationToken);
```
smailliwcs, line 367: Could benefit from a helper method that takes care of serialization and writing to blob storage ... then call into that method here and below, once per blob.  And might be worth looking into whether we can `JsonSerializer.Serialize` _directly_ to the blob stream, rather than first serializing to a string, then converting to a byte array, then writing to the blob stream.

DotNet/Submission/Application/Services/BlobStorageService.cs
```diff
@@ -112,24 +145,548 @@ public async Task UploadToExternalAsync(
...
+            // submitting-org.json - use Organization from manifest
+            if (organization != null)
+            {
+                // Inject dummy address if missing or only contains data-absent-reason extensions
+                bool needsDummyAddress = organization.Address == null ||
+                                         !organization.Address.Any() ||
+                                         organization.Address.All(a =>
+                                             string.IsNullOrEmpty(a.City) &&
+                                             string.IsNullOrEmpty(a.State) &&
+                                             (a.Line == null || !a.Line.Any()));
```
smailliwcs, line 460: Is there a problem with submitting an `Organization` with no address?  If so, can we skip the file entirely?  Or if it's essential, throw an exception (probably earlier in this method)?

DotNet/Submission/Application/Services/BlobStorageService.cs
```diff
@@ -112,24 +145,548 @@ public async Task UploadToExternalAsync(
...
+
+            // Map report types to folder names based on upload_file.py logic
+            if (reportTypesStr.Contains("ach") && reportTypesStr.Contains("hypo"))
+            {
+                return "NHSNdQMAcuteCareHospitalInitialPopulation_NHSNGlycemicControlHypoglycemicInitialPopulation";
+            }
+            else if (reportTypesStr.Contains("rps"))
+            {
+                return "NHSNRespiratoryPathogensSurveillanceInitialPopulation";
+            }
```
smailliwcs, line 623: Should probably revise (even if this matches the existing upload_file.py) to use the "modern" dQM names.

  - NHSNAcuteCareHospitalMonthlyInitialPopulation
  - NHSNAcuteCareHospitalDailyInitialPopulation (instead of RPS)

DotNet/Submission/Application/Services/BlobStorageService.cs
```diff
@@ -112,24 +145,548 @@ public async Task UploadToExternalAsync(
...
+
+            // Fallback: use the first report type or empty string
+            return reportTypes.FirstOrDefault() ?? string.Empty;
+        }
+
+        private HashSet<(string ResourceType, string ResourceId)> FindReferencesInResources(List<Resource> resources)
```
smailliwcs, line 629: This method doesn't seem to be used.  Nor does `FindReferencesInJson`, except by this method.  Do we need them?

smailliwcs (changes requested):

PR #1394, opened 2026-01-28
LNK-4624: DA Worker Queue and Concurrency

edward-miller-lcg:
### 🛠️ Description of Changes
Added a background service, parallel for each, internal channel queue, and immediate kafka commit. Resolved race conditions from introduction of the former. 

### 🧪 Testing Performed
Ran a patient with 8,305 reportable FHIR resources. The report was generated in ~8 minutes.
image

### 🧑‍🔬 Unit Testing

### 📓 Documentation Updated
configs/dotnet/dotnet.mdx

## Summary by CodeRabbit

* **New Features**
  * Introduced asynchronous background processing pipeline for acquisition tasks with configurable concurrency limits.
  * Added a new "Queued" status in the acquisition workflow to better track processing state.

* **Improvements**
  * Enhanced error resilience with automatic retry mechanisms for failed acquisitions.
  * Optimized system performance through bounded work queue management with backpressure control.


DotNet/DataAcquisition.AcquisitionWorker/Listeners/ReadyToAcquireListener.cs
```diff
@@ -47,65 +50,65 @@ protected override ConsumerConfig CreateConsumerConfig()
...
-                    Notes = log.Notes,
-                    Status = log.Status,
-                }, cancellationToken);
-            }
-
-            _logger.LogError(ex, "Error processing ReadyToAcquire message with log id: {LogId}, and facility id: {FacilityId}", consumeResult.Message.Value.LogId, consumeResult.Message.Value.FacilityId);
-            throw new DeadLetterException("Error processing ReadyToAcquire message: " + ex.Message, ex);
+            _logger.LogError(ex, "Failed to enqueue work item for LogId {LogId}. Attempting to revert status.", logId);
+            // Minimally invasive: set back to Pending so the next trigger can try again
+            log.Status = RequestStatus.Pending;
```
amphillipsLGC, line 98: In the enqueue method the try/catch block is only catching the OperationCanceledExceptiontion which would make sense to revert the status to Pending. Is there any chance some other unhandled exception would occur where we would want to actually want to stop trying to reprocess this?

DotNet/DataAcquisition.AcquisitionWorker/Listeners/ReadyToAcquireListener.cs
```diff
@@ -47,65 +50,65 @@ protected override ConsumerConfig CreateConsumerConfig()
...
-            if(log == null)
-            {
-                throw new DeadLetterException($"No DataAcquisitionLog found for log id: {logId}");
-            }
+        // Now we must add the Note (since ExecuteUpdate doesn't handle the Notes list easily)
+        // We can do this as a secondary update because we've already "locked" the record status
+        var log = await logQueries.GetAsync(logId, cancellationToken);
+        log.Notes ??= new List<string>();
+        log.Notes.Add($"[{DateTime.UtcNow:O}] Queued for background acquisition processing");
+        await logManager.UpdateAsync(new UpdateDataAcquisitionLogModel
```
amphillipsLGC: In the above TrySetLogToQueuedAsync method, we are already updating the status and modified date, I think we could add the note as well and not need to make this call and update separately.

DotNet/DataAcquisition.AcquisitionWorker/appsettings.json
```diff
@@ -2,6 +2,8 @@
   "ServiceInformation": {
     "ServiceName": "Link Data Acquisition Worker Service"
   },
+  "AcquisitionWorkerProcessorSettings:MaxConcurrentAcquisitions": 8,
+  "AcquisitionWorkerProcessorSettings:WorkChannelCapacity": 200,
```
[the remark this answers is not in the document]
edward-miller-lcg: documentation added

DotNet/DataAcquisition.AcquisitionWorker/Program.cs
```diff
@@ -23,13 +24,21 @@
 
 var consumerSettings = builder.Configuration.GetRequiredSection(nameof(ConsumerSettings)).Get<ConsumerSettings>();
 
+//register worker processor config
+builder.Services.Configure<AcquisitionWorkerProcessorSettings>(
+    builder.Configuration.GetSection("AcquisitionWorkerProcessorSettings"));
+
```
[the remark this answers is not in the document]
edward-miller-lcg: updated for testing

DotNet/DataAcquisition.AcquisitionWorker/Listeners/ReadyToAcquireListener.cs
edward-miller-lcg: So, there is explicit handling for the timeout since that is concrete evidence that the channel is full and we want to note that. It may be a sign that either 1. Channel isn't big enough or 2. we are not horizontally scaled enough. Other exceptions can/will be handled by the caller. Let me know your thoughts there.

DotNet/DataAcquisition.AcquisitionWorker/Listeners/ReadyToAcquireListener.cs
edward-miller-lcg: After testing a solution to this, I kept the approach as-is due to a need to `get` the log.

DotNet/DataAcquisition.AcquisitionWorker/Listeners/ReadyToAcquireListener.cs
```diff
@@ -47,65 +50,64 @@ protected override ConsumerConfig CreateConsumerConfig()
...
-        DataAcquisitionLogModel? log = null; 
-        _logger.LogInformation("Processing ReadyToAcquire message with log id: {LogId}, and facility id: {FacilityId}", consumeResult.Message.Value.LogId, consumeResult.Message.Value.FacilityId);
+        var logManager = scope.ServiceProvider.GetRequiredService<IDataAcquisitionLogManager>();
+        var processor = scope.ServiceProvider.GetRequiredService<AcquisitionProcessorBackgroundService>();
 
-        try
+        // ATOMIC STEP: Attempt to "claim" the log
+        // This replaces the GetAsync -> Check Status -> UpdateAsync flow
+        var logId = value.LogId.Value;
+        bool claimed = await logQueries.TrySetLogToQueuedAsync(logId, cancellationToken);
```
smailliwcs, line 61: With this addition, do we still need to lock `{facilityId}-{logId}` in `PatientDataService.ExecuteLogRequest`?  If not, can/should we get rid of that lock?

edward-miller-lcg: Technically not. Though, I would think that there is an argument for redundancy. I can attempt to remove that semaphore and see if we maintain behavior with the smoke test and megapatient.

edward-miller-lcg: removed semaphore, retested, same results, resolving.

DotNet/DataAcquisition.AcquisitionWorker/Services/AcquisitionProcessorBackgroundService.cs
```diff
@@ -0,0 +1,165 @@
...
+            }
+
+            // Core business logic - unchanged
+            await patientDataService.ExecuteLogRequest(
+                new AcquisitionRequest(log.Id, item.FacilityId),
+                ct);
+
+            _logger.LogInformation("Successfully completed acquisition for LogId {LogId}", log.Id);
+        }
+        catch (Exception ex) when (ex is not OperationCanceledException)
```
smailliwcs, line 137: - Have you tested that this handles putting any `Queued` logs into `Failed` status during a graceful shutdown of the service?  And that they're picked back up (after a trip through `ReadyToAcquire`) on the next restart?
  - What's our backup plan for a _non_-graceful shutdown?  (I.e., how do we avoid logs getting left in `Queued` status forever?)
  - Do we need the `Queued` status at all?  Couldn't we just do the "atomic status update" thing when moving from `Ready` to `Processing`?
  - How would multiple DA Workers pick up the same log anyway?  Wouldn't that require consuming the same `ReadyToAcquire` message (which shouldn't be possible)?

Not necessarily something to resolve as part of this PR, but do we need a follow-on ticket?

smailliwcs (approved):

DotNet/DataAcquisition.AcquisitionWorker/Services/AcquisitionProcessorBackgroundService.cs
edward-miller-lcg: > What's our backup plan for a non-graceful shutdown? (I.e., how do we avoid logs getting left in Queued status forever?)

This is a good point. I can add a method to the acquisition timed job that checks for logs with a queue longer than .....10 minutes? And, then put that log into a `Failed` status.

> Do we need the Queued status at all? Couldn't we just do the "atomic status update" thing when moving from Ready to Processing?

I added queued because, in my opinion, it's a valid state between Processing and Ready. Ready indicates it's ready to be sent over in kafka, queued means that it's internally, within the worker, to be processed and processed means processed. If you think atomic update is more than enough to cover that middle status, I'm open to removing it.

> Have you tested that this handles putting any Queued logs into Failed status during a graceful shutdown of the service? And that they're picked back up (after a trip through ReadyToAcquire) on the next restart?

I have not. I can test this before merging.

smailliwcs: Agreed that `Queued` makes sense, is good for observability, etc.  But I don't see how to avoid:

  - Either "losing" logs (because they're queued to a worker that subsequently crashes),
  - Or else processing them duplicatively (because we decide they've been queued for too long, and we fail/retry them, but the original worker _hasn't_ actually crashed).

Should a worker be able to process `_channelCapacity` (currently 200) logs in ten minutes?  Probably, but that's out of our control, since it depends on how quickly the EHR services the associated queries.  And I don't think there's a principled way to tune those two quantities such that we're guaranteed to process each log exactly once.

I suppose we could have the worker itself respect the same timeout.  I.e., when it gets ready to move a log from `Queued` to `Processing`, check to see if it's been queued longer than ten minutes, and if so, ignore it (or maybe _fail_ it).  Regardless, if there's a timeout at play, it's probably best for transitions _from_ `Queued` status to be made atomically as well.  That way, if a worker picks up a log right at the time limit, only one attempt to update its status (to either `Processing` or `Failed`) will succeed.

That said, I'm still curious about this:

> How would multiple DA Workers pick up the same log anyway?  Wouldn't that require consuming the same `ReadyToAcquire` message (which shouldn't be possible)?

I.e., does "claiming" logs atomically get us anything that Kafka doesn't already guarantee?  (Equivalently, did we ever really need the mutex that got removed from `PatientDataService`?)

DotNet/DataAcquisition.Domain/Application/Services/PatientDataService.cs
```diff
@@ -310,135 +337,147 @@ public async Task ExecuteLogRequest(AcquisitionRequest request, CancellationToke
...
+            if (!request.ignoreStatusConstraint && log.Status != RequestStatus.Queued)
+            {
+                throw new ArgumentException(
+                    $"Log with ID {log.Id} is not in a queued state. Current status: {log.Status}");
+            }
 
-                //2. set to "Processing"
+            //2. atomically update to "Processing"
+            var successfullyUpdatedLog = await _dataAcquisitionLogQueries.TrySetLogStatusAsync(log.Id,
+                new List<RequestStatus> { RequestStatus.Queued, RequestStatus.Failed }, RequestStatus.Processing,
```
smailliwcs: If the log is `Failed`, we might be racing the DA processing job to see who can update the status first ... either to `Processing` here, or to `Ready`/`MaxRetriesReached` there.  And since the other update isn't atomic, there's a risk of sending the log back through `ReadyToAcquire` even as it's being processed here.  So I'd recommend using `Queued` as the only valid current status.  (Because then the only other update we might be racing, in the watchdog task, is also atomic.)

smailliwcs (approved):

arianamihailescu (commented):
What happens if a record (log)  is committed and then the DAW service goes down before the record being processed or failed? Does it get to be retried?

edward-miller-lcg:
> What happens if a record (log) is committed and then the DAW service goes down before the record being processed or failed? Does it get to be retried?

The Acquisition job will check for stale log records (15 mins) in a queue status and flip them to failed. That way the retry mechanism for ready to acquire will take effect.

DotNet/DataAcquisition.Domain/Application/Services/PatientDataService.cs
edward-miller-lcg: That makes sense. I changed it to just 'queued',

arianamihailescu (commented):
Is Queued the only Status to look for? Can be other statuses to look for as well?

edward-miller-lcg:
> Is Queued the only Status to look for? Can be other statuses to look for as well?

Queued is it, if a node goes down while a log record is queued in the channel, then it is effectively "lost". There isn't anything that will change that status outside of the job checking for those stale records.

PR #1442, opened 2026-02-17
LNK-4830: index changes

edward-miller-lcg:
### 🛠️ Description of Changes

adding indexes for db

### 🧪 Testing Performed

built solution and deployed to local instance with existing db. migration ran fine. e2e was successful for 5 patient and megapaitent.

### 🧑‍🔬 Unit Testing

- Coverage: 0.0%

### 📓 Documentation Updated

none

## Summary by CodeRabbit

* **Chores**
  * Optimized database indexes and applied constraints to improve query performance and data retrieval efficiency for data acquisition logs.


DotNet/DataAcquisition.Domain/Infrastructure/Context/DataAcquisitionDbContext.cs
```diff
@@ -139,17 +139,21 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
                 v => JsonSerializer.Deserialize<ScheduledReport>(v, new JsonSerializerOptions())
             );
 
+            entity.Property(d => d.Status)
+                .HasConversion(new EnumToStringConverter<RequestStatus>())
```
seanmcilvenna, line 143: Suggest updating `RequestStatus` enum to note that enumeration values should not exceed 50 characters (otherwise it will fail to insert/update)

DotNet/DataAcquisition.Domain/Infrastructure/Context/DataAcquisitionDbContext.cs
```diff
@@ -139,17 +139,21 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
+                .HasMaxLength(50);
+
             entity.Property(d => d.Priority)
-                .HasConversion(new EnumToStringConverter<AcquisitionPriority>());
+                .HasConversion(new EnumToStringConverter<AcquisitionPriority>())
```
seanmcilvenna, line 147: Same max length comment suggest as RequestStatus

DotNet/DataAcquisition.Domain/Infrastructure/Entities/DataAcquisitionLog.cs
```diff
@@ -28,6 +28,7 @@ public class DataAcquisitionLog
 
     public QueryPhase? QueryPhase { get; set; }
 
+    [MaxLength(50)]
```
seanmcilvenna, line 31: Does this do compiler-level enforcing of enum values with max lengths? Does this prevent a developer from adding an enum value that is > 50 characters?

seanmcilvenna (changes requested):

DotNet/DataAcquisition.Domain/Infrastructure/Entities/DataAcquisitionLog.cs
edward-miller-lcg: 1. Database Schema Definition
The primary effect is on the database. It tells EF Core to create the column as nvarchar(50) instead of the default nvarchar(max). This is strictly required for the column to be used as a Key Column in a SQL Server index, which was the original goal for optimizing your DataAcquisitionLog queries.
2. Validation Behavior
To your question about preventing assignments:
•
Compile-Time: No. Since the property type is RequestStatus? (the Enum), you can only assign valid members of that Enum (e.g., RequestStatus.Pending) or null. C# will not let you "assign" a string to it directly.
•
Runtime (In-Memory): No. Assigning a value to the property in your code won't trigger a length check automatically.
•
EF Core Saving: Yes (Partially). If you were to add a new Enum member in the future whose name exceeds 50 characters (e.g., RequestStatus.VeryLongStatusNameThatExceedsTheFiftyCharacterLimit), EF Core would attempt to save that string to the database. At that point:
◦
SQL Server will throw an exception (String or binary data would be truncated) because the value won't fit in the nvarchar(50) column.
◦
If you call DbContext.Entry(entity).GetValidationErrors(), EF Core's built-in validation would flag it as an error before even hitting the database.

DotNet/DataAcquisition.Domain/Infrastructure/Context/DataAcquisitionDbContext.cs
edward-miller-lcg: done

DotNet/DataAcquisition.Domain/Infrastructure/Context/DataAcquisitionDbContext.cs
edward-miller-lcg: done

seanmcilvenna (approved):

PR #1446, opened 2026-02-19
LNK-4817: Static scan fixes

johnbritton:
### 🛠️ Description of Changes

Introduced SanitizeUntrustedString extension to clean control characters from untrusted input, preventing log injection and log forging. Refactored logging and ID usage to apply this method throughout the codebase. Added comprehensive unit tests to ensure correct and secure behavior. Improves overall application security.

### 🧪 Testing Performed

Added unit tests for SanitizeUntrustedString

### 🧑‍🔬 Unit Testing

- Coverage: 47.1%

### 📓 Documentation Updated

N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed report schedule ID handling to properly escape special characters in URLs.

* **Tests**
  * Added comprehensive test suite for string sanitization validation.

* **Chores**
  * Enhanced security throughout the application by sanitizing sensitive data in logs before output.


smailliwcs (approved):

DotNet/Shared/Application/Services/Security/StringSecurityExtensions.cs
```diff
@@ -0,0 +1,48 @@
...
+        private static char cleanChar(char aChar)
+        {
+            // This explicitly excludes:
+            // - Control characters (0-31) including tab, newline, carriage return
+            // - DEL character (127)
+            // This must done in a loop so that Fortify can recognize the control character check and not flag this as a potential log injection vulnerability.
+            for (int i = 32; i <= 255; i++)
+            {
+                if (aChar == i && aChar != 127) return (char)i;
+            }
```
smailliwcs, line 44: Approved, but moderately worried about performance.  I assume from the comment that we can't just do the simpler check for `aChar` >= 32, <= 255, and != 127?

edward-miller-lcg (approved):

DotNet/Shared/Application/Services/Security/StringSecurityExtensions.cs
johnbritton: Unfortunately, no. The scan considers anything from the user to be "tainted" no matter how much sanitization you do to it. The only way to make the issue go away is to break that tracking so that it thinks it came from the system, which is what this goofy loop does (while also addressing the real issue).

As for performance, it would only need to be run once on user-entered strings. I expect strings from the user should generally be short.

PR #1447, opened 2026-02-19
LNK-4799: Location Mapping Feature Branch (Do Not Merge)

nvmLantana:
### 🛠️ Description of Changes

Feature branch for the Location Mapping feature.

### 🧪 Testing Performed

### 🧑‍🔬 Unit Testing

- Coverage: 82.6%

### 📓 Documentation Updated

## Release Notes

* **New Features**
  * Added location configuration management capability enabling users to define and manage facility-specific location settings and associated conditions for data acquisition workflows.


DotNet/DataAcquisition.Domain/Application/Queries/OrganizationLocationConfigurationQueries.cs
```diff
@@ -0,0 +1,123 @@
...
+                    FhirPath = cond.FhirPath,
+                    Priority = cond.Priority ?? 1,
+                    CreatedOn = cond.CreatedOn ?? DateTime.UtcNow,
+                    ModifiedOn = cond.ModifiedOn ?? DateTime.UtcNow
+                }).ToList()
+            })
+            .FirstOrDefaultAsync();
+    }
+
+    public async Task<PagedConfigModel<OrganizationLocationConfigurationModel>> SearchAsync(
```
seanmcilvenna, line 82: I doubt we'll need to search across all facilities like this... I *think* the only way we would access these configs is within the context of a specific facility. Let me know if you disagree.

DotNet/DataAcquisition.Domain/Infrastructure/Context/DataAcquisitionDbContext.cs
```diff
@@ -219,6 +221,34 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
...
+
+            entity.HasOne(d => d.Config).WithMany(p => p.LocationConditions)
+                .OnDelete(DeleteBehavior.ClientSetNull)
+                .HasConstraintName("FK_LocationCondition_ConfigId");
+        });
+
+        //-------------------OrganizationLocationConfiguration-------------------
+        modelBuilder.Entity<OrganizationLocationConfiguration>(entity =>
+        {
+            entity.HasKey(e => e.ConfigId).HasName("PK_LocationConfiguration_ConfigId");
```
seanmcilvenna, line 470: Probably need an index of facilityId

DotNet/DataAcquisition.Domain/Infrastructure/Entities/OrganizationLocationCondition.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace LantanaGroup.Link.DataAcquisition.Domain.Infrastructure.Entities;
+
+[Table("OrganizationLocationCondition")]
+[Index("ConfigId", Name = "IX_LocationConditions_ConfigId")]
+public partial class OrganizationLocationCondition
+{
+    [Key]
+    public int ConditionId { get; set; }
+
+    public int ConfigId { get; set; }
```
seanmcilvenna, line 19: Do we need a `[Required]` flag on this? Resolve this if it's covered by the `entity.Property(e => e.FacilityId).Isrequired()` above...

DotNet/DataAcquisition.Domain/Infrastructure/Entities/OrganizationLocationConfiguration.cs
```diff
@@ -0,0 +1,36 @@
...
+    public int ConfigId { get; set; }
+
+    [Required]
+    [StringLength(255)]
+    public string FacilityId { get; set; }
+
+    [StringLength(500)]
+    public string Description { get; set; }
+
+    public bool? IsActive { get; set; }
```
seanmcilvenna: Consider removing the `?` and just let it default to `false`

seanmcilvenna (changes requested):

DotNet/DataAcquisition.Domain/Application/Queries/OrganizationLocationConfigurationQueries.cs
nvmLantana: I am not sure, at this point, what the UI requirements will be, especially for managing the facilities. If we have a page where we want to get a list of all configs with optional filters, facilityId might be good to not require. This question came up a lot when were building out the normalization UI and we ended up with two different routes, one where facilityId is required, and another where it isn't. IMO we should just let this be until we have actual requirements on hand, one way or the other.

DotNet/DataAcquisition.Domain/Infrastructure/Entities/OrganizationLocationCondition.cs
nvmLantana: On what? There is no FacilityId on the OrganizatonLocationCondition table. There is a ConfigId, which is not nullable and is enforced via the FK constraint (0 won't match an existing OrganizationLocationConfiguration)

DotNet/DataAcquisition.Domain/Infrastructure/Entities/OrganizationLocationConfiguration.cs
nvmLantana: Good suggestion, implemented.

DotNet/DataAcquisition.Domain/Infrastructure/Context/DataAcquisitionDbContext.cs
nvmLantana: There already is such an index, but it is not setup in the DbContext. It is setup in the migration scripts, which create it in the database, and also annotated on the OrganizationLocationConfiguration entity.

DotNet/DataAcquisition.Domain/Application/Queries/OrganizationLocationConfigurationQueries.cs
nvmLantana: In grooming, we decided to require the FacilityId on Search. Implemented.

DotNet/DataAcquisition.Domain/Extensions/GeneralStartupExtensions.cs
```diff
@@ -241,6 +245,7 @@ public static void RegisterManagers(this IServiceCollection services)
         services.AddTransient<IReferenceResourcesManager, ReferenceResourcesManager>();
         services.AddTransient<IFhirQueryManager, FhirQueryManager>();
         services.AddTransient<IDataAcquisitionLogManager, DataAcquisitionLogManager>();
+        services.AddScoped<IOrganizationLocationConfigurationManager, OrganizationLocationConfigurationManager>();
```
smailliwcs, line 309: Preexisting queries/manager dependencies are transient.  Should the new ones be transient as well.  Alternatively, should the old ones be scoped?

DotNet/DataAcquisition.Domain/Infrastructure/Entities/OrganizationLocationCondition.cs
```diff
@@ -0,0 +1,36 @@
...
+
+    [Required]
+    [StringLength(500)]
```
smailliwcs: I worry that a really hairy condition could exceed this.  E.g., what if there are 20 possible `Location.identifier` values that map to a particular OrgID, and we want to OR all of those checks together in this `FhirPath`?  Though maybe we'd handle that by breaking those checks into separate entities?  Regardless, is it worth considering bumping the maximum length up to 1000?  Or handling this property as an `nvarchar(max)`?

DotNet/DataAcquisition.Domain/Infrastructure/Entities/OrganizationLocationCondition.cs
```diff
@@ -0,0 +1,36 @@
...
+    public string FhirPath { get; set; }
+
+    public int? Priority { get; set; }
```
smailliwcs: Seems like we're using 1 as the default priority elsewhere.  Can we assign that here and make the property non-nullable?

On that note, is priority 1 intended to be high or low?  What do you think about making 0 the default priority, and we could explicitly denote "above nominal" priority as a positive value, "below nominal" as negative?

DotNet/DataAcquisition.Domain/Infrastructure/Entities/OrganizationLocationCondition.cs
nvmLantana: Yeah I think Nvarchar(max) will work. The original size you see here was before the full FhirPath leveraging. Good shout.

DotNet/DataAcquisition.Domain/Extensions/GeneralStartupExtensions.cs
nvmLantana: The old ones should be scoped, but I did not want to make that change as part of this feature.

DotNet/DataAcquisition.Domain/Infrastructure/Entities/OrganizationLocationCondition.cs
nvmLantana: Priority is meant to be read as 'order of condition analysis', not one condition being more important than the other. Defaulted to 1, not nullable anymore.

DotNet/DataAcquisition.Domain/Infrastructure/Entities/OrganizationLocationCondition.cs
nvmLantana: implemented nvarchar(max)

smailliwcs (approved):

PR #1464, opened 2026-02-26
LNK-4857: Resolve Data Acq Job Deadlocks

edward-miller-lcg:
### 🛠️ Description of Changes

This PR implements a series of high-impact performance optimizations and stability fixes for the Data Acquisition service to resolve consistent SQL deadlocks and execution timeouts observed in high-volume environments.
•
High-Speed Bulk Updates: Refactored status transitions in DataAcquisitionLogManager and DataAcquisitionLogQueries to use EF Core's ExecuteUpdateAsync. This moves the update logic to the database server, replacing slow "fetch-then-update" cycles with single, set-based SQL commands.
•
LINQ Translation Resolution: Removed appending failure notes to the JSON Notes collection during batch operations. This change was necessary because EF Core 7+ cannot translate C# collection methods (like .Append().ToList()) into SQL for JSON-mapped columns, which previously caused runtime exceptions and forced a fallback to inefficient row-by-row updates.
•
Database Optimization:
◦
Added composite indexes on (Status, ModifyDate) and (Status, ExecutionDate) to the DataAcquisitionLog table.
◦
These indexes enable the AcquisitionProcessingJob and cleanup tasks to perform index seeks instead of full table scans, drastically reducing the duration and scope of database locks.
•
Kafka Key Handling: Updated DataAcquisitionRequestedListener to correctly handle Kafka message keys formatted as JSON (e.g., ResourceKey). It now extracts the raw facilityId from the JSON object, preventing configuration retrieval failures.
•
EF Core Stability: Added ValueComparers for complex JSON properties (QueryPlan.InitialQueries, FhirListConfiguration.EHRPatientLists, etc.) in DataAcquisitionDbContext to resolve "missing value comparer" warnings and ensure accurate change tracking.
•
Job Efficiency: Ensured all "work discovery" queries use .AsNoTracking() to minimize memory overhead and prevents unnecessary object tracking during parallel job execution.

### 🧪 Testing Performed

Testing was successfully completed in the LCG TEST environment. Verified that the AcquisitionProcessingJob can process large volumes of logs concurrently without triggering deadlocks or hit the 30-second SQL command timeout.

### 🧑‍🔬 Unit Testing

- Coverage: 0.0%
(Updated DataAcquisitionLogManagerTests to support constructor dependency changes).

### 📓 Documentation Updated

n/a

## Summary by CodeRabbit

* **New Features**
  - Configurable data acquisition job scheduling interval via application settings (default: every 10 seconds)

* **Improvements**
  - Enhanced reliability with automatic retry handling for transient database deadlocks
  - Optimized query performance for read-only operations


DotNet/DataAcquisition.Domain/Application/Managers/DataAcquisitionLogManager.cs
```diff
@@ -199,16 +203,18 @@ public async Task<List<DataAcquisitionLog>> GetLogsByIdsAsync(List<long> ids, Ca
...
-        foreach (var log in logs)
-        {
-            log.ModifyDate = DateTime.UtcNow;
-            _database.DataAcquisitionLogRepository.Update(log);
-        }
-        await _database.DataAcquisitionLogRepository.SaveChangesAsync(cancellationToken);
+        using var activity = ServiceActivitySource.Instance.StartActivity("DataAcquisitionLogManager.UpdateStatusBatchAsync");
+
+        // High-speed bulk update without fetching entities. 
+        // Note: The 'note' parameter is currently ignored to maintain performance and avoid LINQ translation errors with JSON collections.
```
smailliwcs: If we can't update notes here, I'd recommend removing the parameter to avoid confusion in calling code.

DotNet/DataAcquisition.Domain/Application/Queries/DataAcquisitionLogQueries.cs
```diff
@@ -99,13 +102,22 @@ public class DataAcquisitionLogQueries : IDataAcquisitionLogQueries
...
     public DataAcquisitionLogQueries(IDatabase database, DataAcquisitionDbContext dbContext,
         ILogger<DataAcquisitionLogQueries> logger)
     {
         _database = database ?? throw new ArgumentNullException(nameof(database));
         _dbContext = dbContext ?? throw new ArgumentNullException(nameof(dbContext));
         _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+
+        _deadlockRetryPolicy = Policy
+            .Handle<SqlException>(ex => ex.Number == 1205) // SQL Deadlock error number
+            .WaitAndRetryAsync(5, retryAttempt => TimeSpan.FromMilliseconds(Math.Pow(2, retryAttempt) * 100),
```
smailliwcs: Recommend adding a random component to the delay so that we don't re-deadlock on each retry.  I'm seeing stuff in the Polly docs related to `UseJitter`, which I think would achieve what we want.

DotNet/DataAcquisition.Domain/Application/Queries/DataAcquisitionLogQueries.cs
```diff
@@ -399,52 +417,18 @@ public async Task<bool> TrySetLogToQueuedAsync(long logId, CancellationToken can
...
-            catch (Exception ex)
-            {
-                await transaction.RollbackAsync(cancellationToken);
-                _logger.LogError(ex, "Failed to process a batch of stalled logs. Error: {Message}", ex.Message);
-                throw;
-            }
-        }
 
-        if (batchesProcessed >= maxBatches)
+        // High-speed bulk update without fetching entities or using transactions.
```
smailliwcs, line 423: I think this needs to remain transactional, otherwise we could be failing logs while some DA Worker is simultaneously picking them up for processing.

DotNet/DataAcquisition/appsettings.json
```diff
@@ -76,6 +76,9 @@
     "OtelCollectorEndpoint": "",
     "EnableAzureMonitor": false
   },
+  "AcquisitionJobSettings": {
+    "CronSchedule": "0/10 * * * * ?"
```
smailliwcs, line 80: Why the discrepancy with the `0/30` value in code?  (If we're setting it in code, do we need a value here?)  And does this need to be called out in `docs/docs/custom/config/dotnet.mdx`?

smailliwcs (changes requested):

DotNet/DataAcquisition.Domain/Application/Managers/DataAcquisitionLogManager.cs
edward-miller-lcg: done

DotNet/DataAcquisition.Domain/Application/Queries/DataAcquisitionLogQueries.cs
edward-miller-lcg: complete.

DotNet/DataAcquisition.Domain/Application/Queries/DataAcquisitionLogQueries.cs
edward-miller-lcg: good point. reverted to transactional approach.

DotNet/DataAcquisition/appsettings.json
edward-miller-lcg: I made it configurable so hence why there is a default AND an appsettings.json setting. I'll get a PR over to link-ec as well.

edward-miller-lcg: Commit for documentation: https://github.com/lantanagroup/link-ec/commit/715ea723aca54d754cbdca41cf7c8f17194d3cb5

DotNet/DataAcquisition.Domain/Application/Queries/DataAcquisitionLogQueries.cs
```diff
@@ -418,16 +420,38 @@ public async Task<int> FailStalledQueuedLogsAsync(int stallMinutes, int maxBatch
...
-        // This avoids LINQ translation errors with JSON collections and minimizes lock duration.
         return await _deadlockRetryPolicy.ExecuteAsync(async () =>
         {
-            return await _dbContext.DataAcquisitionLogs
-                .Where(l => l.Status == RequestStatus.Queued && l.ModifyDate <= stallThreshold)
-                .ExecuteUpdateAsync(setters => setters
-                    .SetProperty(l => l.Status, RequestStatus.Failed)
-                    .SetProperty(l => l.ModifyDate, DateTime.UtcNow),
-                    cancellationToken);
+            using var transaction = await _dbContext.Database.BeginTransactionAsync(cancellationToken);
```
smailliwcs: Ugh, I see now that I was wrong about needing an explicit transaction here.  The previous strategy of running a single `ExecuteUpdateAsync` is probably preferable.  Recommend reverting `7d224fe` ... apologies for the rework.

smailliwcs (approved):

DotNet/DataAcquisition.Domain/Application/Queries/DataAcquisitionLogQueries.cs
edward-miller-lcg: no problem! re-added.

PR #1468, opened 2026-03-02
LNK-4857: Fix Op Outcome retries

edward-miller-lcg:
### 🛠️ Description of Changes

Ensure that logs with OperationOutcome results are not retried.
image

### 🧪 Testing Performed
Local E2E with Single Measure Adhoc report with 5 patients as well as a 251 patient report with multiple patients with multiple operation outcomes as well as megapatient.

### 🧑‍🔬 Unit Testing

- Coverage: 0.0%

### 📓 Documentation Updated

n/a

## Summary by CodeRabbit

* **New Features**
  * Added "Inoperable" status for data acquisition logs to indicate operations that cannot be processed.
  * Enhanced FHIR API error handling with detailed outcome tracking and logging.
  * UI now supports filtering and displaying the new "Inoperable" status.

* **Chores**
  * Updated NuGet package dependencies to latest compatible versions.


smailliwcs (commented):
Not sure I understand the need for the new `RequestStatus`.  If we get an OO in a 404 or 410 response, shouldn't we call that `Completed` since it's a "successfully empty" result?  And if it's some other status code, shouldn't we handle it the same as any other `Failed` request?

More to the point, if we're mainly trying to resolve the issue of endless retries of 404'd patient reads, wouldn't removing the `fhirQuery.IsReference` check [here](https://github.com/lantanagroup/link-cloud/pull/1468/changes#diff-a56d23c582f9a3395ed525be12e03857849f20bd5cce43c95f1c315dc4b423e3R142) achieve that?

DotNet/DataAcquisition.Domain/Application/Services/PatientDataService.cs
```diff
@@ -632,7 +665,7 @@ await _dataAcquisitionLogQueries.UpdateAsync(new UpdateDataAcquisitionLogModel
             log.RetryAttempts ??= 0;
 
             log.ExecutionDate = DateTime.UtcNow.Add(ex.RetryAfter);
-            log.Status = RequestStatus.Pending; //Don't count this as a failure
+            log.Status = RequestStatus.Failed; //Don't count this as a failure
```
smailliwcs, line 668: Was this change supposed to be made on line 620?  Looks like this also impacts the release branch.

smailliwcs (changes requested):

dvargaslantana (approved):

DotNet/DataAcquisition.Domain/Application/Services/PatientDataService.cs
edward-miller-lcg: fix pushed here. will validate in release.

edward-miller-lcg:
> Not sure I understand the need for the new `RequestStatus`. If we get an OO in a 404 or 410 response, shouldn't we call that `Completed` since it's a "successfully empty" result? And if it's some other status code, shouldn't we handle it the same as any other `Failed` request?
> 
> More to the point, if we're mainly trying to resolve the issue of endless retries of 404'd patient reads, wouldn't removing the `fhirQuery.IsReference` check [here](https://github.com/lantanagroup/link-cloud/pull/1468/changes#diff-a56d23c582f9a3395ed525be12e03857849f20bd5cce43c95f1c315dc4b423e3R142) achieve that?

Removed the new status and added better http status handling.

PR #1469, opened 2026-03-02
LNK-4857: Fix Op Outcome retries (RELEASE BRANCH)

edward-miller-lcg:
### 🛠️ Description of Changes

Ensure that logs with OperationOutcome results are not retried.
image

### 🧪 Testing Performed
Local E2E with Single Measure Adhoc report with 5 patients as well as a 251 patient report with multiple patients with multiple operation outcomes as well as megapatient.

### 🧑‍🔬 Unit Testing

- Coverage: 0.0%

### 📓 Documentation Updated

n/a

## Release Notes

* **New Features**
  * Added "Inoperable" status for data acquisition logs to better handle FHIR operation errors.
  * Enhanced error handling with improved logging for FHIR operations.
  * UI now displays the new status with appropriate styling.

* **Tests**
  * Expanded test coverage for error scenarios and status handling.

* **Chores**
  * Updated package dependencies to latest versions.

### DIFF
--- a/DotNet/DataAcquisition.Domain/Application/Queries/DataAcquisitionLogQueries.cs
+++ b/DotNet/DataAcquisition.Domain/Application/Queries/DataAcquisitionLogQueries.cs
@@ -202,7 +202,7 @@ public class DataAcquisitionLogQueries : IDataAcquisitionLogQueries
                 where l.FacilityId == facilityId
                       && l.ReportTrackingId == reportTrackingId
                       && l.CorrelationId == correlationId
-                      && !(l.Status == RequestStatus.Completed || l.Status == RequestStatus.MaxRetriesReached)
+                      && !(l.Status == RequestStatus.Completed || l.Status == RequestStatus.MaxRetriesReached || l.Status == RequestStatus.Inoperable)
                       && !l.TailSent
                       && l.FhirQueries.Any(fq => fq.IsReference == false)
                 select l).CountAsync(cancellationToken);
@@ -213,7 +213,7 @@ public class DataAcquisitionLogQueries : IDataAcquisitionLogQueries
         CancellationToken cancellationToken = default)
     {
         var completedOrFailedStatuses = new[]
-            { RequestStatus.Completed, RequestStatus.MaxRetriesReached, RequestStatus.Skipped };
+            { RequestStatus.Completed, RequestStatus.MaxRetriesReached, RequestStatus.Skipped, RequestStatus.Inoperable };
 
         try
         {
--- a/DotNet/DataAcquisition.Domain/Application/Queries/DataAcquisitionLogQueries.cs
+++ b/DotNet/DataAcquisition.Domain/Application/Queries/DataAcquisitionLogQueries.cs
@@ -202,7 +202,7 @@ public class DataAcquisitionLogQueries : IDataAcquisitionLogQueries
                 where l.FacilityId == facilityId
                       && l.ReportTrackingId == reportTrackingId
                       && l.CorrelationId == correlationId
-                      && !(l.Status == RequestStatus.Completed || l.Status == RequestStatus.MaxRetriesReached || l.Status == RequestStatus.Inoperable)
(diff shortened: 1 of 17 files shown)


DotNet/DataAcquisition.Domain/Application/Queries/DataAcquisitionLogQueries.cs
```diff
@@ -202,7 +202,7 @@ public async Task<int> GetCountOfNonRefLogsIncompleteAsync(string facilityId, st
                 where l.FacilityId == facilityId
                       && l.ReportTrackingId == reportTrackingId
                       && l.CorrelationId == correlationId
-                      && !(l.Status == RequestStatus.Completed || l.Status == RequestStatus.MaxRetriesReached)
+                      && !(l.Status == RequestStatus.Completed || l.Status == RequestStatus.MaxRetriesReached || l.Status == RequestStatus.Inoperable)
```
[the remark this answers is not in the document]
edward-miller-lcg: removed inoperable status and operation outcome results are marked with status of completed.

DotNet/DataAcquisition.Domain/Application/Services/PatientDataService.cs
```diff
@@ -601,6 +601,27 @@ await _dataAcquisitionLogQueries.UpdateAsync(new UpdateDataAcquisitionLogModel
                 }, cancellationToken);
             }
         }
+        catch (OpOutcomeException ex)
+        {
+            _logger.LogWarning(ex, "OperationOutcome encountered for facility {FacilityId}", log.FacilityId.Sanitize());
+
+            log.Notes ??= new List<string>();
+            log.Status = RequestStatus.Completed;
```
smailliwcs: If I understand the changes to `FhirApiService` correctly, this makes it so that any 404/410 or OO-bearing response causes the log to be marked as `Completed`?  But aren't there status codes that we _want_ to retry, even if there's an OO in the response?  E.g., what if a network glitch results in a 401/403/500/503, and the response happens to have an OO in the body?  Don't we still want to mark the log as `Failed` in that case?

Does the logic in `FhirApiService` need to care about the presence/absence of an OO at all?

  - If 404/410, the EHR is saying, "I looked and there's nothing there."  OK, log can be `Completed`.
  - If any other `FhirOperationException`, then something-bad-but-we-don't-know-what happened.  Log should be `Failed` for retry?

smailliwcs (changes requested):

DotNet/DataAcquisition.Domain/Application/Services/PatientDataService.cs
edward-miller-lcg: ok, added logic that only marks 404 and 410 as completed, otherwise it will be failed and go into retry.

DotNet/DataAcquisition.Domain/Application/Services/PatientDataService.cs
```diff
@@ -601,6 +602,38 @@ await _dataAcquisitionLogQueries.UpdateAsync(new UpdateDataAcquisitionLogModel
...
+            if (ex.StatusCode is HttpStatusCode.NotFound or HttpStatusCode.Gone)
+            {
+                log.Status = RequestStatus.Completed;
+                log.CompletionDate = DateTime.UtcNow;
+            }
+            else
+            {
+                log.RetryAttempts ??= 0;
+                log.RetryAttempts++;
+                log.Status = RequestStatus.Pending;
```
smailliwcs, line 620: Does the `AcquisitionProcessingJob` handle this correctly?  Specifically, since we're skipping `Failed` and going straight to `Pending`, if we keep getting an `OpOutcomeException` for this log, will it be retried endlessly?  Or will something eventually (correctly) set it to `MaxRetriesReached`?

smailliwcs (approved):

DotNet/DataAcquisition.Domain/Application/Services/PatientDataService.cs
edward-miller-lcg: Yes, it's supposed to be Failed. That was an oversight on my part and have corrected it. Thank you.

PR #1498, opened 2026-03-13
TECH_DEBT: Added exception handling for null blob storage containers when attempting to upload

dvargaslantana:
### 🛠️ Description of Changes

 Added exception handling for null blob storage containers when attempting to upload.

### 🧪 Testing Performed

Tested to make sure the exception is properly thrown when there is a null container

image

## Release Notes

* **Bug Fixes**
  * Enhanced error handling in blob storage operations to throw exceptions when the storage client is unavailable, improving visibility into storage-related failures during file uploads and manifest operations.

### 🧑‍🔬 Unit Testing
- Coverage: 0.0%


nvmLantana (approved):
Looks good.

PR #1579, opened 2026-04-21
LNK-5140: Restore Reference Resource Query Efficiency

nvmLantana:
### 🛠️ Description of Changes

# LNK-5140: First-class, batched, cache-aware reference resource acquisition

## Summary

This PR changes how reference resources are acquired in Data Acquisition. On `dev`, reference resources are fetched inline as a side effect of primary log execution: one FHIR `Read` per referenced resource, not visible as their own acquisition logs, not honoring the reference query plan configuration, and re-fetching the same resources across correlations. This PR makes reference acquisition a first-class, batched, cache-aware activity that goes through the normal acquisition pipeline.

### Reference acquisition is now a first-class `DataAcquisitionLog`

- Exactly one durable reference log per `(FacilityId, CorrelationId, QueryPhase, ResourceType)`.
- Created at the end of primary log execution from the references discovered in that primary's bundles.
- Visible, retryable, and auditable like any other acquisition log, instead of being an invisible side effect.

### Reference fetches honor the query plan configuration

The reference query plan's settings are now actually used when fetching:

- operation type (`Search` vs `SearchPost`)
- `Paged` batch size

On `dev` these were effectively ignored for reference fetches.

### Batched `_id=...` searches instead of one-by-one `Read`s

- All references discovered for a given resource type within a correlation are unioned and deduplicated into a single reference log.
- That log issues batched `_id=a,b,c,...` searches respecting the configured batch size.
- This replaces the dev behavior of issuing one FHIR `Read` per referenced resource.

### Canonical facility-level reference cache reuse

- `ReferenceResources` is consulted before any FHIR call.
- Cache hits are linked to the executing reference log and still produce `ResourceAcquired` events for the active correlation, so downstream consumers see them as part of this run.
- Cache misses are fetched once via the batched search above, persisted canonically, linked to the log, and produce `ResourceAcquired` events.
- Subsequent correlations for the same facility do not refetch resources already in the cache.

### Correct sequencing within a correlation

- A reference log becomes eligible only after all non-reference siblings in the same `(facility, correlation, phase)` are terminal, so the full set of needed references is known before fetching starts.
- Tail completion waits for the reference log before downstream signaling fires, so downstream services do not start evaluating a correlation that is still missing reference resources.

### Concurrency-safe reference log creation

- Creation of a reference log within a `(facility, correlation, phase)` is serialized through a single app lock.
- The new reference log stamps only its own row; no wide cross-row updates participate in this path.
- Tail completion derives its threshold from the live count of stamped siblings, so reference logs added mid-flow are accounted for without contending with other group-wide writes.

### Cleanup

- Removed the `Referential` query phase and related shape from the shared model and Data Acquisition.
- Removed the inline reference-fetch path and helper models that no longer apply.
- Trimmed comments in touched areas.

## Schema

- Adds `DataAcquisitionLog.ReferenceResourceType` and a unique key for `(FacilityId, CorrelationId, QueryPhase, ReferenceResourceType)` so there is exactly one reference log per group.
- Establishes `PendingReferenceIds` as a per-reference-log ID bag (FK to `DataAcquisitionLog`, unique on `(DataAcquisitionLogId, ResourceId)`).

## User-visible outcomes

- Reference acquisition is visible per correlation and per resource type as durable acquisition log entries.
- The FHIR server is hit far less for references: at most one batched search per resource type per correlation, and no fetch at all when the canonical cache already has the resource.
- Reference work is retried, audited, and reasoned about the same way as primary work.
- Multi-measure scenarios behave the same as single-measure scenarios with respect to reference acquisition.

## Suggested reviewer order

1. `Shared/Application/Models/Integration/DataAcquisition/DataAcquisitionApiModels.cs`
2. `DataAcquisition.Domain/Infrastructure/Entities/*` and `Infrastructure/Context/DataAcquisitionDbContext.cs`
3. New migration under `DataAcquisition.Domain/Migrations/`
4. `ReferenceResourceService`
5. `PatientDataService`
6. `FhirApiService`
7. `DataAcquisitionLogQueries` and `DataAcquisitionLogManager`

## Release Notes

* **New Features**
  * Implemented two-phase reference resource acquisition: references are now staged during initial processing and promoted to dedicated referential logs once the initial phase completes, improving efficiency and reducing contention.
  * Added automatic batched FHIR Search capability for reference resources, restoring optimized bulk retrieval behavior.

* **Bug Fixes**
  * Fixed reference resource acquisition to use batched queries instead of individual requests, improving performance.

* **Tests**
  * Added comprehensive integration tests for referential phase promotion and staging behavior.

### 🧑‍🔬 Unit Testing
- Coverage: 3.1%


amphillipsLGC (approved):
Tested locally and verified the reference resources were showing up in the acquisition logs.

PR #1606, opened 2026-04-29
LEGLINK-45: Update Automation To Offer External FHIR Data Uploads

nvmLantana:
### 🛠️ Description of Changes

Updated Automation.UI so that it can intake patient json or an existing PatientId to reference on the fhir server and integrate the data into a test scenario, including Report time framing, Predictive modeling, FHIR server data management, etc.

Fixed the Data Acquisition Log Filtering by providing Search Term support in the LogQueries in DataAcquisition.

Removed PatientId prefixing, as it was a meaningless holdover from earlier designs.

Performed some campsite cleanup from previous designs.

Updated PR Title Check to Accept LEGLINK-XX.

Fixed multiple miscenllaneous bugs encountered during design and testing.

Updated ReadMe documents with changes.

### 🧪 Testing Performed

Ran all System reports, tested Patient Upload and Patient Lookup test scenarios, with and without cohort generation.

### 🧑‍🔬 Unit Testing

- Coverage: 34.3%

### 📓 Documentation Updated

## Release Notes

* **New Features**
  * Added free-text search for data acquisition logs with debouncing and server-side filtering
  * Added support for importing existing FHIR patients by ID or transaction bundle
  * Added configurable report period start/end dates for scenarios

* **Improvements**
  * Streamlined scenario configuration by removing patient ID prefix mechanism
  * Enhanced quick-launch form to mirror saved scenario values

* **Removals**
  * Removed legacy patient ID prefix functionality


DotNet/Automation/Generation/ImportedPatientClassifier.cs
```diff
@@ -0,0 +1,122 @@
...
+        return new ClassificationResult(elig, detectedScenarioId);
+    }
+
+    private static bool IsInpatient(Encounter enc)
+    {
+        var code = enc?.Class?.Code;
+        if (string.IsNullOrWhiteSpace(code)) return false;
+        return string.Equals(code, "IMP", StringComparison.OrdinalIgnoreCase)
+               || string.Equals(code, "ACUTE", StringComparison.OrdinalIgnoreCase)
+               || string.Equals(code, "NONAC", StringComparison.OrdinalIgnoreCase);
```
smailliwcs: This is "correct" in that it detects the three ActEncounterCodes that represent inpatient encounters.  However, an encounter does _not_ need to be inpatient in order to qualify for the IP.  Specifically, in addition to these three inpatient codes, `EMER`, `OBSENC`, and `SS` would also qualify.  Furthermore, even if the `class` is some other code (or nonexistent), the encounter _could_ still qualify for the IP via its `type`, or via an associated location's `type`.

Long story short, we're not accurately replicating the CQL's IP inclusion logic.  But given how complex that logic is, I'm not sure we want to try.  Granted, "not trying" amounts to not being able to predict Q/NQ, but as it stands, that prediction won't be accurate in general.  (I'm guessing it has worked up to this point because Automation's factories are generating data in a way that happens to work for its prediction logic.)

DotNet/Automation/Generation/ImportedPatientClassifier.cs
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    private static bool IsDiabeticHypoglycemicCondition(Condition cond)
```
smailliwcs, line 148: Same comment here.  While Automation's factories might always put a hypoglycemia `Condition` in a qualifying patient's bundle, that's not technically required to be in the hypo IP.  What _is_ required is an antidiabetic drug request/administration during an IP encounter, which again is probably more complex of a situation than we want to try to detect.

DotNet/Automation/Generation/MeasureInitialPopulationResolver.cs
```diff
@@ -0,0 +1,186 @@
...
+            ProfiledMeasureType.NhsnAcuteCareHospitalDailyInitialPopulation =>
+                MatchesClass(cls, "IMP", "EMER", "AMB"),
+
+            // Hypoglycemic: inpatient encounter only.
+            ProfiledMeasureType.NhsnGlycemicControlHypoglycemicInitialPopulation =>
+                MatchesClass(cls, "IMP"),
+
+            _ => false
+        };
+    }
```
smailliwcs: Ditto again.  Also, these codes are different from the ones in `ImportedPatientClassifier`.

smailliwcs (approved):

DotNet/Automation/Generation/ImportedPatientClassifier.cs
nvmLantana: Thank you for these comments, I will take this information and refine the prediction logic. I will keep pushing this concept forward until I reach a point where it becomes untenable. I think, for an automation tool, the end result will be suffiicent.

PR #1644, opened 2026-06-03
LEGLINK-65: Add PostSearch for Parameter queries

MikeAtPinnacle:
### 🛠️ Description of Changes

Added the ability to select a query type when entering a Parameter query.  Available options are Search and PostSearch

### 🧪 Testing Performed

Changed the Encounter parameter to a PostSearch.  Ran a report.  Observed PostSearch and Search was used in the Acquisition Logs when the appropriate query type is used.

### 🧑‍🔬 Unit Testing

- [X] I have written or updated unit tests to cover my changes
- Coverage: 100.0%

### 📓 Documentation Updated

## Summary by CodeRabbit

* **New Features**
  * Operation Type can be set for parameter queries; UI exposes the field and conditionally hides "Read" except for reference queries. Paged field visibility updated.

* **Improvements**
  * Validation now rejects invalid or forbidden operation types for parameter queries.
  * SearchPost (POST-based) search path supported.

* **Tests**
  * New unit tests covering operation-type handling, validation, and POST search behavior.

### DIFF
--- a/DotNet/DataAcquisition.Domain/Application/Factories/QueryFactories/ParameterQueryFactory.cs
+++ b/DotNet/DataAcquisition.Domain/Application/Factories/QueryFactories/ParameterQueryFactory.cs
@@ -6,6 +6,7 @@ using LantanaGroup.Link.DataAcquisition.Domain.Infrastructure.Models.QueryConfig
 using LantanaGroup.Link.DataAcquisition.Domain.Infrastructure.Models.QueryConfig.Parameter;
 using LantanaGroup.Link.Shared.Application.Models;
 using Microsoft.Extensions.Logging;
+using StackExchange.Redis;
 using OperationType = LantanaGroup.Link.DataAcquisition.Domain.Infrastructure.Models.QueryConfig.OperationType;
 
 namespace LantanaGroup.Link.DataAcquisition.Domain.Application.Factories.QueryFactories;
@@ -84,7 +85,13 @@ public class ParameterQueryFactory : IParameterQueryFactory
             }
         }
 
-        return isPaged ? new PagedParameterQueryFactoryResult(OperationType.Search, searchParamList)
-            : new SingularParameterQueryFactoryResult(OperationType.Search, searchParams);
+        // Read is not a valid operation type for a parameter query, 
+        // but if it is set to that, treat it as a search to avoid errors and still return results
+        var operationType = config.OperationType == OperationType.SearchPost
+         ? OperationType.SearchPost 
+         : OperationType.Search;
+
+        return isPaged ? new PagedParameterQueryFactoryResult(operationType, searchParamList)
+            : new SingularParameterQueryFactoryResult(operationType, searchParams);
     }
 }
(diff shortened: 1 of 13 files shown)


DotNet/DataAcquisition.Domain/Application/Factories/QueryFactories/ParameterQueryFactory.cs
```diff
@@ -84,7 +85,13 @@ public async Task<ParameterQueryFactoryResult> Build(ParameterQueryConfig config
             }
         }
 
-        return isPaged ? new PagedParameterQueryFactoryResult(OperationType.Search, searchParamList)
-            : new SingularParameterQueryFactoryResult(OperationType.Search, searchParams);
+        // Read is not a valid operation type for a parameter query,
```
seanmcilvenna, line 88: If set to Read in this case, maybe log a warning?

DotNet/ServiceTests/UnitTests/DataAcquisition/Factories/ParameterQueryFactoryTests.cs
```diff
@@ -144,5 +144,61 @@ public async Task Build_WhenParameterFactoryReturnsNull_SkipsIt()
...
+                    new LiteralParameter { Name = "lit1", Literal = "val1" }
+                }
+            };
+            var request = new GetPatientDataRequest { CorrelationId = "corr1", FacilityId = "fac1" };
+
+            // Act
+            var result = await _parameterQueryFactory.Build(config, request, null, null, _dataAcquisitionLogQueriesMock.Object);
+
+            // Assert
+            var singularResult = Assert.IsType<SingularParameterQueryFactoryResult>(result);
```
seanmcilvenna, line 170: I think the asserts here should go deeper than just the `.ofType` test.
When performing a standard Search it should result in a query `GET /type?param1=value1&param2=value2`
But when performing a Search with POST style, it should be `POST /type/_search` and the `body` would include `param1=value1&param2=value2`
I think this should check that difference between query parameters in Search vs. body in SearchPost.

seanmcilvenna (changes requested):

DotNet/ServiceTests/UnitTests/DataAcquisition/Factories/ParameterQueryFactoryTests.cs
mtherien: Would you then be testing the FHIR library not this code?  The purpose of this is to make sure the right method on the FHIR library was called.

MikeAtPinnacle: Looking at this specific test, this is making sure the new logic is setting the correct operation type for the PagedParameterQueryFactoryResult.  What you mention I cover in a different test.

MikeAtPinnacle: I think you are referring to this test: https://github.com/lantanagroup/link-cloud/blob/users/mtherien/LEGLINK-65-post-search/DotNet/ServiceTests/UnitTests/DataAcquisition/SearchFhirCommandTests.cs#L333-L354

Since we don't have a FhirClientFactory to mock, this was the next best thing.  I can create that so the test is very specific.

DotNet/DataAcquisition.Domain/Application/Factories/QueryFactories/ParameterQueryFactory.cs
MikeAtPinnacle: Fixed

seanmcilvenna (approved):

PR #1647, opened 2026-06-04
LEGLINK-355: Query Plan Update

nvmLantana:
### 🛠️ Description of Changes

https://lantana.atlassian.net/browse/LEGLINK-355

Update & Fix The Query Plan Builder

Update the Query Plan Simulator and ensure Custom Query Plan selection is used in prediction.

### 🧪 Testing Performed

Ran locally and observed proper functioning of Query Plan UI and successful test results.

### 🧑‍🔬 Unit Testing

- Coverage: 30.6%

### 📓 Documentation Updated

### DIFF
--- a/DotNet/Automation/Generation/QueryPlanAcquisitionSimulator.cs
+++ b/DotNet/Automation/Generation/QueryPlanAcquisitionSimulator.cs
@@ -151,14 +151,20 @@ public static class QueryPlanAcquisitionSimulator
                 }
             }
 
-            if (string.Equals(p.ParameterType, "Literal", StringComparison.OrdinalIgnoreCase)
-                && string.Equals(p.Name, "category", StringComparison.OrdinalIgnoreCase))
+            if (string.Equals(p.ParameterType, "Literal", StringComparison.OrdinalIgnoreCase))
             {
-                if (!MatchesLiteralCategory(resource.Resource, p.Literal))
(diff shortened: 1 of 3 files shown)


DotNet/Automation/Generation/QueryPlanAcquisitionSimulator.cs
```diff
@@ -283,36 +289,126 @@ private static bool TryGetReferencedResourceId(JsonElement resource, string reso
...
+
+        if (string.Equals(paramName, "recorded-date", StringComparison.OrdinalIgnoreCase))
+        {
+            return TryGetInstant(resource, "recordedDate", out start, out end);
+        }
+
+        // Generic "date" → resource-type-aware resolution.
+        return TryGetResourceDateRange(resourceType, resource, out start, out end);
+    }
+
```
smailliwcs, line 411: Probably not an issue for us, but wanted to note that `paramName` might not uniquely identify a FHIR search parameter ... the same name might refer to multiple SPs that correspond to different resource types and even have different data types.  E.g., when the `start` parameter is used for the Slot resource type, it's a date representing the appointment date/time, but when used for the GraphDefinition resource type, it's a token (string) representing the resource type at which the graph starts.

But I don't think that affects any of the SPs we're currently using, so we'll likely never need to worry about it.

smailliwcs (approved):

PR #1683, opened 2026-06-23
LEGLINK-516: Automation.UI Do not persist full ABS Data

nvmLantana:
### 🛠️ Description of Changes

in Automation.UI, do not store scenario run ABS data. This data was used for diagnostics, post run, which is a rare case. 

Instead, store lightweight meta data about the ABS submission, and if diagnostics are run, retrieve the ABS data if it is still present for the diagnostic export. 

### 🧪 Testing Performed

### 🧑‍🔬 Unit Testing

- Coverage: 0.0%

### 📓 Documentation Updated

## Summary by CodeRabbit

* **New Features**
  * Export process now supports re-downloading archived artifacts with fallback to cached snapshots, improving data freshness and reliability.

* **Bug Fixes**
  * Fixed tooltip formatting in scenario editor.
  * Enhanced report period boundary detection for imported patient encounters.


DotNet/Automation.UI/Services/RunExportService.cs
```diff
@@ -359,36 +361,142 @@ private static string BuildNormalizationSection(
...
+                    continue;
+
+                using var entryStream = entry.Open();
+                using var reader = new StreamReader(entryStream);
+                files[entry.FullName] = await reader.ReadToEndAsync(ct);
+            }
+
+            return files;
+        }
+        catch (Exception ex)
```
smailliwcs, line 455: If `ex` is an `OperationCanceledException`, just rethrow?  Otherwise it's treated as a download failure and the caller keeps plugging away.

DotNet/Automation/Generation/AbsExportLocatorSnapshot.cs
```diff
@@ -0,0 +1,44 @@
...
+            .OrderBy(k => k, StringComparer.OrdinalIgnoreCase)
+            .ToList();
+
+        long charCount = 0;
+        foreach (var value in files.Values)
+        {
+            charCount += value switch
+            {
+                string s => s.Length,
+                _ => value?.ToString()?.Length ?? 0
```
smailliwcs: If `files` typically contains Firely `Resource`s, wouldn't we need to deserialize here (instead of just calling `ToString`) to get the character count?

smailliwcs (approved):

PR #1697, opened 2026-07-01
LEGLINK-541: Ingest CodeSystem status column and add cached-code lookup endpoint

MikeAtPinnacle:
### 🛠️ Description of Changes

Adds support for an optional **status** column in CodeSystem CSV ingestion and a diagnostic endpoint to inspect cached codes.

- CodeSystem CSVs may now have **2 or 3 columns** (`code`, `display`, and optional `status`). Adds the `CodeStatus` enum (`Active`/`Inactive`) and a `CodeSystemCode` model that carries the status.
- Status defaults to **Active** when the column is absent (2-column CSV) or the cell is empty. The CSV reader is configured with `MissingFieldFound = null` so a missing trailing column is not treated as an error.
- `CodeStatus` serializes as its name (`"Active"`/`"Inactive"`) via `JsonStringEnumConverter` rather than its numeric value.
- New diagnostic endpoint: `GET api/terminology/config/CodeSystem/{id}/code/{code}` returns a single cached code (value, display, status) looked up by CodeSystem resource id. Inputs are sanitized; returns 400 on missing id/code and 404 when the CodeSystem or code is not cached.

### 🧪 Testing Performed

- Unit + LoadCache-level tests in `CodeGroupCacheServiceTests` (all passing).
- Manual verification against the local `link-terminology` container: confirmed the `v3-ActCode` CodeSystem CSV ingests as a 3-column file, and the new endpoint returns e.g. `{ "value": "DIET", "display": "Diet", "status": "Active" }`.  (I manually modified the `c3-ActCode` file to add the status column to test)

### 🧑‍🔬 Unit Testing

- Coverage: 51.3%

### 📓 Documentation Updated

No changes required.

## Summary by CodeRabbit

* **New Features**
  * Added support for code statuses, including active and inactive values.
  * Code-system imports can now include an optional status column, with missing or blank values defaulting to active.
  * Added a new lookup endpoint to retrieve a specific code from a code system by ID and version.

* **Bug Fixes**
  * Improved code-system loading to handle CSV files with either 2 or 3 columns more reliably.
  * Expanded test coverage for cache loading and code parsing behavior.


DotNet/Terminology/Application/Models/CodeSystemCode.cs
smailliwcs, line 1: It's unclear from the diff, but I assume there's a reason we need a new `CodeSystemCode` subclass, as opposed to just putting the `Status` property in the preexisting `Code`?  Does active/inactive status apply _only_ to codes in `CodeSystem`s, and _not_ to codes in `ValueSet`s?

smailliwcs (approved):

DotNet/Terminology/Application/Models/CodeSystemCode.cs
MikeAtPinnacle: @smailliwcs Well, I didn't like the name I picked, but that is irrelevant.  Yes, the status applies only to the code system, not to value sets.

PR #1753, opened 2026-07-22
LEGLINK-582: Validation service writes category-grouped pre-qualification OperationOutcome to patient NDJSON

MikeAtPinnacle:
### 🛠️ Description of Changes

The Validation service now writes the pre-qualification `OperationOutcome` into the patient NDJSON in ABS, becoming the sole writer of it. Report's flat "Patient has failed Validation" append is retired separately in LEGLINK-466.

**OperationOutcome construction** (`PreQualOperationOutcomeBuilder`, new)
- Builds a single `OperationOutcome` with **one issue per unacceptable category** — a category with `acceptable=false` in `categories.json` that has at least one finding. Acceptable categories are excluded. A `Result` mapping to several categories contributes an issue to each.
- Each issue carries the `pq-issue-cat` extension (category id), `severity=error`, `code=processing`, and the first finding's message as `details.text`.
- The `OperationOutcome` carries the `oo-total` extension (total issue count).
- Expressions anchor to the patient's MeasureReport (`Bundle.entry[0].resource.ofType(MeasureReport).where(id = '…')`) followed by each finding's own location expression.
- A patient with no unacceptable-category findings produces **no** `OperationOutcome`, and nothing is written.

**Blob write** (`BlobStorageService`)
- Was download-only; added `appendResource`, appending one NDJSON line to the existing patient append blob via `AppendBlobClient.appendBlock`. It does not call `create()` — the blob is created upstream by Report's `PatientAggregator.AggregateToABS`.
- The appended line is newline-terminated so a subsequent appender cannot concatenate onto the same line. This was caught during manual testing: without it, Report's still-active legacy OperationOutcome landed on the *same* line and produced malformed NDJSON. It also protects against the two runtimes' flags drifting out of sync.

**Wiring** (`ReadyForValidationConsumer`)
- After categorization, resolves the blob name from the `ReadyForValidation` `payloadUri` (the same value already used for download), extracts the MeasureReport id from the in-hand bundle, builds the `OperationOutcome`, and appends it.
- Fully flag- and null-guarded, so a stack without blob storage configured (local/dev) is a no-op.

**Configuration** (`PreQualificationConfig`, new)
- `pre-qualification.write-pre-qual-operation-outcome` (default `false`) gates the entire write.
- `pre-qualification.write-expressions-in-operation-outcome` (default `true`) gates `issue.expression[]`.
- These are the Java-side keys for a cross-runtime concept; the .NET Report service reads the equivalent `PreQualification:WritePreQualOperationOutcome` key in LEGLINK-466. Both bind from Azure App Config through the existing `bootstrap.yml` wiring, are registered in `app-config.yaml`, and are exposed as `docker-compose` environment overrides.

**Testing aid**
- New Postman collection (`Tests/Postman/leglink-425-prequal-operationoutcome.postman_collection.json`) that ensures a test facility with query plan and FHIR query config, seeds a patient into the local HAPI server, and triggers an ad-hoc report to drive the whole pipeline. All variables are defined in-collection, so it needs no Postman environment. Setup requests are idempotent.

**Behavior on merge is unchanged** — the master flag defaults to `false` everywhere, so nothing is written until it is explicitly enabled.

### 🧪 Testing Performed

**Unit tests** — `mvn -pl validation -am test`, 47 tests pass (0 failures, 0 errors) on JDK 17. This grew from 37 over the review round below.

**Review fixes** — a review pass raised four latent defects and two coverage gaps, all addressed on this branch. Each defect was masked by an incidental runtime property rather than being reachable on the happy path: the MeasureReport locator hard-coded `Bundle.entry[0]` and only worked because the aggregator writes it first; issue grouping was keyed by the `Category` entity, which declares no `equals`/`hashCode`, and only worked because Hibernate returns one instance per id per session; the blob append was not replay-safe (see LEGLINK-800 below); and record-derived values reached the logger unsanitized. The coverage gaps were the `payloadUri == null` branch of the append guard and `appendResource` itself, which had no direct test pinning the trailing newline that a real defect had already required.

**Manual end-to-end against the local docker-compose stack:**
1. Enabled `pre-qualification.write-pre-qual-operation-outcome=true` on the validation container and rebuilt the image.
2. Seeded a facility, query plan, FHIR query config and a patient via the new Postman collection, then triggered an ad-hoc report for `NHSNAcuteCareHospitalMonthlyInitialPopulation`.
3. Downloaded the resulting `patient-<id>.ndjson` from Azurite and confirmed the appended `OperationOutcome`:
   - `oo-total` = 3, matching the three issues emitted.
   - Each issue carried the correct `pq-issue-cat` category id, `severity=error`, `code=processing`, and the category's message.
   - Expressions resolved against the live MeasureReport id, confirming MeasureReport-id extraction works on a real aggregated bundle.
4. This run surfaced the missing-newline defect described above; after the fix the appended resource terminates its own line.

**Flag behavior verified** by unit test rather than a second full pipeline run: append occurs only when the flag is set, is skipped when there are no unacceptable-category findings, and is skipped when blob storage is unavailable.

**Not yet exercised:** the `missing_active_*` (inactive-code) categories ship as `acceptable=true`, so inactive-code findings are tolerated and produce no issue. Confirming them requires temporarily flipping those categories, which is a content decision rather than a code change and was deliberately not included here.

### 🧑‍🔬 Unit Testing

- Coverage: 98.9%
  - **`PreQualOperationOutcomeBuilderTest` (new, 14 tests)** — one issue per unacceptable category; `oo-total` equals issue count; `pq-issue-cat`, severity, code and `details.text` content; first-result message selected per category; a single result mapping to multiple unacceptable categories yielding an issue each; acceptable categories excluded; empty result when nothing is unacceptable; null/empty category lists ignored; expressions present with the MeasureReport locator first when enabled and absent entirely when disabled; null MeasureReport id omitting the locator without failing; MeasureReport-id extraction (present, absent, null bundle).
  - **`ReadyForValidationConsumerTest` (+4 tests)** — appends to the same blob when the flag is on and unacceptable findings exist; does not append when the flag is off, when there are no unacceptable findings, or when blob storage is absent.
  - `BlobStorageService.appendResource` is not unit tested directly, since it is a thin wrapper over the Azure SDK client; it is covered indirectly through the mocked consumer tests, consistent with how the existing `download` method is treated.

### 📓 Documentation Updated

- `app-config.yaml` — both new configuration keys registered under the `Validation` service, including a note that `write-pre-qual-operation-outcome` is the Java side of a cross-runtime flag whose .NET counterpart arrives with LEGLINK-466.
- `docker-compose.yml` — both keys exposed as environment overrides on the validation service, defaulting to off.
- The new Postman collection is self-documenting: each request describes its purpose and preconditions, including that `categories.json` is a database seed baked into the image (so exercising the inactive-code categories requires a rebuild plus `POST /api/validation/category/$initialize`), and which codes can be reported inactive in a default local stack.
- No changes required to the Admin UI: the `sub-pre-qual-report` view consumes the JSON pre-qualification summary shape, not the `pq-issue-cat` / `oo-total` FHIR extensions.

### 🔗 Related

- Parent story: LEGLINK-425
- LEGLINK-466 gates Report's legacy OperationOutcome write behind the .NET counterpart of this flag. Until it merges, enabling this flag results in **both** OperationOutcomes being appended. The trailing newline added here keeps the NDJSON well-formed in that interim state.
- **LEGLINK-800** tracks full idempotency across the blob and database writes. Review raised that the blob append is not replay-safe: `AsyncListener` acknowledges the Kafka offset even when `process()` throws, so a failure after the append routes the record to the retry topic and it is processed again. This PR adds a replay guard that skips the append when the bundle already carries a pre-qualification OperationOutcome, which narrows the window but does not close it — the REST fallback supplies a bundle with no previously appended OperationOutcome so the check is blind, concurrent consumers can both observe it as absent, and the check and the append are not atomic. Closing it properly needs a durable idempotency marker (new entity, repository and schema migration), and the writes span three systems with no distributed transaction, so no ordering makes them atomic. LEGLINK-800 also covers the pre-existing duplicate `Result` rows the same replay produces via `resultRepository.saveAll`.

## Summary by CodeRabbit

* **New Features**
  * Added optional pre-qualification `OperationOutcome` output to patient NDJSON files.
  * Added configuration controls for enabling output and including issue expressions.
  * OperationOutcome entries now group unacceptable findings by category and include relevant report locations and expressions.
  * Added safe append support for existing patient data files, including duplicate-write prevention.
* **Tests**
  * Added automated and end-to-end coverage for configuration, outcome generation, blob appending, and validation workflows.

### DIFF
--- a/Java/validation/src/main/java/com/lantanagroup/link/validation/configs/PreQualificationConfig.java
+++ b/Java/validation/src/main/java/com/lantanagroup/link/validation/configs/PreQualificationConfig.java
@@ -1,0 +1,28 @@
+package com.lantanagroup.link.validation.configs;
+
+import lombok.Getter;
+import lombok.Setter;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * Flags controlling the pre-qualification OperationOutcome that the Validation service writes to the
+ * patient NDJSON. These are the Java-runtime keys for a cross-runtime concept; the .NET Report service
+ * reads the equivalent {@code PreQualification:WritePreQualOperationOutcome} key (LEGLINK-466).
+ */
+@Getter
+@Setter
+@Configuration
+@ConfigurationProperties("pre-qualification")
+public class PreQualificationConfig {
+    /**
+     * Master gate. When true, the Validation service is the sole writer of the pre-qualification
+     * OperationOutcome to the patient NDJSON. When false (default), it writes nothing.
+     */
+    private boolean writePreQualOperationOutcome = false;
+
+    /**
+     * When false, the {@code expression[]} of each issue is omitted from the OperationOutcome.
+     */
+    private boolean writeExpressionsInOperationOutcome = true;
+}
--- a/Java/validation/src/main/java/com/lantanagroup/link/validation/services/BlobStorageService.java
+++ b/Java/validation/src/main/java/com/lantanagroup/link/validation/services/BlobStorageService.java
@@ -5,6 +5,10 @@ import com.azure.storage.blob.BlobClient;
 import com.azure.storage.blob.BlobContainerClient;
 import com.azure.storage.blob.BlobServiceClient;
 import com.azure.storage.blob.BlobServiceClientBuilder;
+import com.azure.storage.blob.specialized.AppendBlobClient;
+
+import java.io.ByteArrayInputStream;
+import java.nio.charset.StandardCharsets;
 
 public class BlobStorageService {
     private final BlobContainerClient containerClient;
@@ -19,5 +23,20 @@ public class BlobStorageService {
     public BinaryData download(String blobName) {
         BlobClient client = containerClient.getBlobClient(blobName);
         return client.downloadContent();
+    }
+
+    /**
+     * Appends a single NDJSON line (one serialized FHIR resource) to an existing patient NDJSON append
+     * blob. The blob is created upstream by the Report service's aggregation as an append blob, so this
+     * does not call {@code create()}.
+     * <p>
+     * A trailing newline is written so the appended resource always terminates its own line. Without it,
+     * any subsequent appender (e.g. the Report service's legacy OperationOutcome write, which is only
+     * retired once LEGLINK-466's flag is set) concatenates onto this line and produces malformed NDJSON.
+     */
+    public void appendResource(String blobName, String ndjsonLine) {
+        AppendBlobClient client = containerClient.getBlobClient(blobName).getAppendBlobClient();
+        byte[] bytes = (ndjsonLine + "\n").getBytes(StandardCharsets.UTF_8);
+        client.appendBlock(new ByteArrayInputStream(bytes), bytes.length);
     }
 }
--- a/Java/validation/src/main/java/com/lantanagroup/link/validation/services/BlobStorageService.java
+++ b/Java/validation/src/main/java/com/lantanagroup/link/validation/services/BlobStorageService.java
@@ -20,6 +20,14 @@ public class BlobStorageService {
         containerClient = serviceClient.getBlobContainerClient(blobContainerName);
     }
 
+    /**
+     * Visible for testing: takes an already-built container client so the append behaviour can be
+     * exercised against mocks without constructing a real Azure client.
+     */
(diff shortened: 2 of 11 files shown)


arianamihailescu (approved):
Looks good

PR #1770, opened 2026-07-29
LEGLINK-857: Improvements to avoid Redis timeouts

johnbritton:
### 🛠️ Description of Changes

1. Parse Redis connection strings to allow options to be set through configuration rather than hardcoded.
2. Change the distributed semaphore and ResourceCache to use async calls.
3. Connect to Redis using a connection pool instead of a connection multiplexer to avoid client timeouts.

### 🧪 Testing Performed

Unit tests and local integration tests.

### 🧑‍🔬 Unit Testing

- Coverage: 56.0%

### 📓 Documentation Updated

N/A

## Summary by CodeRabbit

* **New Features**
  * Added configurable Redis connection-pool sizing for distributed locks and resource caching, defaulting to five connections.
  * Added support for Redis connection options in Docker deployments to improve connection handling and DNS resolution.
* **Improvements**
  * Cache reads, updates, and deletions now run asynchronously and honor cancellation, improving responsiveness during data acquisition and normalization.
  * Redis-backed caching and distributed locking now use pooled connections.
* **Bug Fixes**
  * Improved cache health checks with clearer failure handling and prioritized Redis validation.


DotNet/Account/appsettings.Docker.json
```diff
@@ -13,7 +13,7 @@
   "DatabaseProvider": "SqlServer",
   "AutoMigrate": true,
   "ConnectionStrings": {
-    "Redis": "redis_cache:6379"
+    "Redis": "redis_cache:6379,abortConnect=false,resolveDns=true,allowAdmin=true"
```
kissalk, line 16: Not sure if it matters, but Managed Redis uses port `10000`

DotNet/Shared/Application/Services/ResourceCache/RedisResourceCache.cs
```diff
@@ -14,23 +16,20 @@ public class RedisResourceCache : IResourceCache
         private readonly IDatabase _db;
         private readonly ILogger<RedisResourceCache> _logger;
 
-        public RedisResourceCache(IConnectionMultiplexer redis, ILogger<RedisResourceCache> logger)
+        public RedisResourceCache(IRedisDatabase redisDatabase, ILogger<RedisResourceCache> logger)
         {
-            _db = redis.GetDatabase();
+            _db = redisDatabase.Database;
```
MikeAtPinnacle: The connection pool doesn't pool — both consumers pin one connection at startup

This is the PR's headline change (#3 in the description), and as wired it's a no-op.

StackExchange.Redis.Extensions' IRedisDatabase.Database is a computed property — I reflected over RedisDatabase in the 11.0.0 assembly and it has no Database backing field; the getter (58 IL bytes) calls connectionPoolManager.GetConnection() on every access. That per-access call is the entire pooling mechanism.

Both consumers evaluate it exactly once and cache the result in a singleton:

- DotNet/Shared/Application/Services/ResourceCache/RedisResourceCache.cs:21 — _db = redisDatabase.Database; in the constructor of a singleton.
- DotNet/Shared/Application/Models/Configs/DistributedLockSettings.cs:109 — new RedisDistributedSynchronizationProvider(database.Database) inside AddSingleton.

So PoolSize = 5 yields a pool where the resource cache permanently uses one connection and the semaphore permanently uses one connection. Under the load that motivated the ticket (the acquire loop firing HashSetAsync per resource), that's the same multiplexer saturation as before, just with four idle spare connections. HybridResourceCache and ResourceCacheHealthCheck re-read .Database per call and do rotate correctly — which makes the contrast a good sanity check.

Fix is to hold IRedisDatabase and read .Database at each call site rather than caching IDatabase.

(found by claude)

app-config.yaml
MikeAtPinnacle: Is this now an actual connection string?  Before it was just the host/port

arianamihailescu (approved):
Looks good

MikeAtPinnacle (changes requested):

DotNet/Shared/Application/Services/ResourceCache/RedisResourceCache.cs
johnbritton: Good catch. I think I have fixed this with https://github.com/lantanagroup/link-cloud/pull/1770/commits/d1ee1faeb6c2bcf0cc9123ce4dba20f6c1c4251e

app-config.yaml
johnbritton: Yes. I updated the file to indicate this.

MikeAtPinnacle (approved):

PR #1774, opened 2026-07-31
LEGLINK-774: Reconcile the App Configuration catalog and gate it in CI

MikeAtPinnacle:
### 🛠️ Description of Changes

`app-config.yaml` is meant to be the authoritative catalog of every configuration key the
platform reads. It had drifted: it listed keys no code reads, omitted keys code does read, and
nothing validated any of it. This PR makes the catalog accurate, adds the three environment
store exports alongside it, and puts CI gates in place so it cannot drift again.

**No changes are made to Azure.** The exports under `Config/` describe intended state.
LEGLINK-775 imports them.

#### The catalog

- 131 entries, up from 118. 88 marked `required: true`, each verified to have a row in all
  three environment stores.
- 34 entries marked `sensitive: true`, up from **zero** — the field was declared in the schema
  and never used. It is now derived from a fact in the stores rather than from judgement: a row
  whose `content_type` is the `keyvaultref` type is that environment declaring the value a
  secret.
- 15 Key Vault backed keys added that were absent entirely — SMTP credentials, Kafka SASL,
  `Redis:Password`, `Logging:HmacKey`, the Admin BFF OpenID Connect trio, and three Java-side
  keys.
- `consumers` populated on 43 of 45 `global` entries, from the `link:consumers` tags the
  exports carry. It only carries information under `global`; under `services` the section
  already names the consumer.
- `serviceMeta` added: each service's App Configuration label and runtime, so tooling reads the
  mapping instead of hardcoding it.

#### The environment exports

`Config/app-config.{dev,qa,test}.json` are committed, with 638 rows tagged with their consuming
services so the Azure portal shows who reads what. Tags are inert — neither provider selects on
them, so they cannot change what any service resolves.

#### CI gates

Two workflows on every PR, plus a pre-commit hook:

- **App Config Secret Scan** — fails if an export carries a credential, or a Key Vault reference
  whose `content_type` is wrong (App Configuration serves those as a literal JSON string
  instead of resolving them, so the service reads `{"uri": "..."}` as its password).
- **App Config Catalog Check** — every `required: true` key has a row in every store; no store
  row carries a label no service selects; `sensitive` and Key Vault backing agree in both
  directions.

Neither filters on paths, deliberately, so both stay valid as required status checks.

#### Tooling

`Scripts/AzureAppConfig/` — 13 scripts with their own README. A Roslyn symbol dump plus a
call-site scanner derive the key inventory from source; a reconciler reports where the catalog,
the code and the stores disagree.

### Defects found and fixed along the way

- **qa was misconfigured** — Automation.UI pointed at `localhost:8063`, Validation had an empty
  terminology service URL.
- **Terminology logged to Loki with `component=""`** — it selects a label that no row carried.
- **`export-appconfigs.bat` silently dropped three labels** (`DataAcquisitionWorker`,
  `Link Automation UI`, `Terminology`) by iterating a hardcoded list. An export produced by it
  and re-imported with `--strict` would have deleted 28+ rows from dev alone.
- **`TelemetrySettings.EnableOtelCollector` is a public field, not a property.**
  `ConfigurationBinder` binds properties only, so the value set for it in all three stores can
  never take effect.
- **A plaintext Slack webhook** in all three live stores. The committed export replaces it with
  a Key Vault reference; the live value needs rotating, tracked separately.

### Known findings recorded but not fixed here

- Three Key Vault backed rows exist in dev under label `Link Automation UI`
  (`Authentication:Schemas:OpenIdConnect:{Authority,ClientId,ClientSecret}`). Repo-wide,
  `OpenIdConnect` is read only by `DotNet/Admin.BFF/`, which selects a different label, so
  nothing consumes them. Deleting them is a store change and belongs to LEGLINK-775.
- The `<Service>:<Environment>` label tier is dead:
  `ExternalConfigurationExtension.cs:64` concatenates the environment object rather than its
  name. No store uses such a label, and the catalog check now rejects any label containing `:`,
  so the trap is enforced shut without a code change.

### 🧪 Testing Performed

- **Catalog checks green against the real data**: schema validator, required-key check and
  secret scanner (`--strict`) all exit 0 against all three environment exports.
- **Every check proven to fail as well as pass.** The rules report clean on the real files, so
  each was run against deliberately mutated fixtures: a required key deleted from one store, a
  label containing `:`, a label absent from `serviceMeta`, an entry with `sensitive: true` and a
  `defaultValue`, a Key Vault row downgraded to a literal, and a Key Vault row with its catalog
  entry removed. Each failed naming the exact key.
- **Refactors verified by output diffing.** The five scripts were captured on both their clean
  and failure paths before the consolidation and folder move, and re-run after: all nine
  byte-identical with unchanged exit codes.
- **Catalog edits verified structurally, not by eye.** Each catalog change was checked by
  parsing `HEAD` and the edited file and walking both trees, asserting that entries were touched
  only in the intended way and that the `services` and `serviceMeta` sections were untouched.
- **Pre-commit hook tested end to end** on Windows — exits 1 against a poisoned export, 0
  against a clean one.
- **Java key notation confirmed by disassembly** rather than assumption: the Spring Cloud Azure
  provider appends `*` to the key filter, strips the leading `/` and converts the rest to `.`,
  so `/spring/datasource/url` resolves as `spring.datasource.url`. Two agents disagreed on this;
  `javap` settled it.

### 🧑‍🔬 Unit Testing

- Coverage: 0.0%
  asserts both the satisfied and unsatisfied direction, per notation class — Java dotted/slash,
  array elements, `{Placeholder}` templates, JSON blob flattening, label scoping, and the
  sensitive/Key Vault rules. No production code changed in this PR, so there is no service test
  coverage to report.

### 📓 Documentation Updated

- `Config/README.md` — the exports, label semantics, and the notation rules.
- `Scripts/README.md` — an inventory of all 43 scripts in the folder, grouped by purpose.
- `Scripts/AzureAppConfig/README.md` — the tooling in detail, including why the gate and the
  reconciler both exist.
- `docs/config-key-inventory.md` — the complete generated list of every key the code reads.
- `docs/design-appconfig.html` — dual dotted/slash notation, label resolution, and that App
  Configuration outranks environment variables in both runtimes.
- `CLAUDE.md` — configuration registry section (local only, not committed).

### Reviewer notes

- **`@lantanagroup/devops` and `@lantanagroup/link-release-management-team`** are code owners of
  `Config/app-config*.json` and `/app-config.yaml`.
- **Needs a repo admin after merge**: add *Validate catalog and required keys* as a required
  status check. It has been enforcing rather than advisory since the gaps it found were closed.
- The bulk of the line count is `docs/config-key-inventory.md` (1,939 lines, generated) and the
  three store exports. The reviewable surface is `app-config.yaml` and `Scripts/AzureAppConfig/`.

## Summary by CodeRabbit

* **New Features**
  * Added complete development, QA, and test App Configuration exports.
  * Added configuration inventory and comparison tools to identify missing, unused, or environment-specific settings.
  * Added automatic consumer tagging for configuration entries.

* **Bug Fixes**
  * Improved validation for required settings, labels, defaults, schema structure, and sensitive values.
  * Added detection of exposed credentials, malformed Key Vault references, duplicate entries, and invalid endpoints.

* **Documentation**
  * Expanded configuration and script guidance, including export, import, validation, labeling, security, and runtime behavior.


Config/app-config.dev.json
```diff
@@ -0,0 +1,2650 @@
...
+			"content_type": "",
+			"tags": {}
+		},
+		{
+			"key": "/spring/kafka/properties/security/protocol",
+			"value": "SASL_PLAINTEXT",
+			"label": null,
+			"content_type": "",
+			"tags": {}
+		},
```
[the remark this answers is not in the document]
MikeAtPinnacle: From claude:

This finding is not safe to apply, and applying it would take the platform down.

Why
1. The brokers only speak SASL_PLAINTEXT. [docker-compose.yml:350](vscode-webview://07msbnlo6oh1iteu9disoajp3p9pvpjge3p79f7pgk1ucfmud2b1/docker-compose.yml#L350) maps every listener to plaintext:


KAFKA_CFG_LISTENER_SECURITY_PROTOCOL_MAP:
  "CONTROLLER:SASL_PLAINTEXT,SASL_PLAINTEXT:SASL_PLAINTEXT,EXTERNAL:SASL_PLAINTEXT"
There is no TLS listener. The deployed clusters are Strimzi (kafka-kafka-bootstrap.kafka-<env>) on port 9092 — Strimzi's plain listener; TLS is 9093. A SASL_SSL client against a plaintext listener fails the handshake outright. Every .NET service and both Java services would lose Kafka. Total pipeline outage in all three environments.

The decisive proof is empirical: the current config works today. If the brokers required SSL, nothing would be running.

2. The client has no TLS configuration surface. [KafkaConnection.cs](vscode-webview://07msbnlo6oh1iteu9disoajp3p9pvpjge3p79f7pgk1ucfmud2b1/DotNet/Shared/Application/Models/Configs/KafkaConnection.cs#L52) sets only SecurityProtocol, SaslUsername, SaslPassword, SaslMechanism. Repo-wide there is no SslCaLocation, ssl.truststore, or certificate setting anywhere. Switching the value wouldn't just need a broker change — it needs code.

3. CodeRabbit's own caveat is satisfied. It said "unless cluster-only access is explicitly guaranteed." It is: kafka-kafka-bootstrap.kafka-dev is a Kubernetes ClusterIP service DNS name that doesn't resolve outside the cluster. This is pod-to-pod traffic on the cluster network, not an exposed endpoint.

4. It would breach this story's contract. These JSON files are exports of actual store state, and LEGLINK-775 imports them. Editing them here means 775 pushes SASL_SSL into the live stores — an infrastructure change delivered through a documentation ticket, with no broker change to match it.

.githooks/pre-commit
```diff
@@ -0,0 +1,85 @@
...
+}
+
+if [ -n "$staged_exports" ]; then
+    files=""
+    for path in $staged_exports; do
+        files="$files $(stage_blob "$path")"
+    done
+
+    # shellcheck disable=SC2086
+    if ! "$PYTHON" Scripts/AzureAppConfig/validate_aac_secrets.py $files; then
```
arianamihailescu: Should we enforce "strict" here on a pre-commit like we do on the workflow "appconfig-secret-scan.yml" that runs on the  CI pipeline?

app-config.yaml
```diff
@@ -15,19 +15,119 @@ properties:
       type: "array"
```
arianamihailescu, line 15: CORS entries do not seem to be captured in this file:
 CORS:AllowCredentials and CORS:AllowAllOrigins

arianamihailescu (commented):
Has this already been removed from the test store? 
CORS:AllowedOrigins:5 = https://qa-admin.nhsnlink.org/    label: LinkAdminBFF

When will this be removed?

.githooks/pre-commit
MikeAtPinnacle: I will make that change

app-config.yaml
MikeAtPinnacle: I added the values, but the purpose of the catalog is to document what can or should be changed in each environment.  Some of these values are the same in each environment and don't need to change.  Not everything in appsettings.json is documented in the catalog.

Config/app-config.test.json
```diff
@@ -0,0 +1,2574 @@
+{
+	"items": [
+		{
+			"key": ".appconfig.featureflag/NotificationSendEmailFeature",
```
amphillipsLGC, line 4: We don't need to change this as I don't think it will cause problems, but I don't think this is currently being used as the Notification service was removed until a future time when it needs to be added back.

amphillipsLGC (approved):

arianamihailescu (approved):
Looks good

PR #1800, opened 2026-08-07
LEGLINK-822: Add Mock DMRP API for reporting-plan integration testing

MikeAtPinnacle:
### 🛠️ Description of Changes

Adds `MockDmrpApi`, a stand-in for the third-party DMRP API that Leidos is building for CDC, so local development and the LCG environments can exercise reporting-plan integration without reaching the CDC Dev network.

**⚠️ The single most important thing for a reviewer to know:** this service hosts **two surfaces**, and only one of them is DMRP.

| | **Contract surface** | **Support surface** |
|---|---|---|
| Routes | `GET /msc`, `GET /ps/annual` | everything under `/mock` |
| Whose API? | The third party's | Ours |
| In `Contracts/dmrp-openapi.yaml`? | Yes — it describes *only* these | No, deliberately |
| Authentication | The third party's bearer token | Link's standard scheme (`IsLinkAdmin`) |
| Exists in real DMRP? | Yes | **No** |

A consumer integrates against the contract surface only. The support surface exists so a test can set up an exact scenario, and it is kept out of the contract so that replacing that document with Leidos's published one cannot disturb it.

**Contract-first with NSwag**

- `Contracts/dmrp-openapi.yaml` drives codegen at build time into `obj/NSwag/`, which is **gitignored** — the C# and the spec cannot drift, and a contract change reviews as a spec diff rather than thousands of lines of mechanical churn.
- The request parameters and response body come from the DMRP Integration ADR, so they match what LCG has recorded of the real API.
- Replacing the spec when Leidos publishes theirs is a wholesale file swap: rebuild, and the compile errors enumerate the work.

**Persistence**

- `MockDmrpEntries` in a new `link-mock-dmrp` catalog, EF Core, with an `InitMockDmrp` migration verified to apply, roll back and re-apply.
- Natural key is `(facilityId, component, reportingYear, reportingMonth, measure)`, unique.
- ⚠️ The index sets `HasFilter(null)`. EF's default for a unique index over a nullable column is `WHERE [ReportingMonth] IS NOT NULL`, which would drop every annual row out of the index and silently permit duplicate patient-safety entries.

**Support surface**

CRUD, filtered search, an OAuth2-shaped token endpoint that issues the credential the contract endpoints accept, and configurable response-delay endpoints for timeout and retry testing.

**Availability**

`MockDmrpApi:Enabled` disables the service, and **production never serves it whatever configuration says**. Azure App Configuration is appended last in the chain, so a row provisioned against a production label would silently outrank appsettings and environment variables — and a running mock looks exactly like a healthy service, so that failure would be invisible. When disabled, every route answers 503, `/health` and `/api/mock-dmrp/info` keep answering, and migration is skipped.

**Also included**

- `docker-compose.yml` on port 6159, a CD pipeline, `app-config.yaml` entries, and a regenerated config-key inventory.
- Errors are RFC 9457 problem details in the same house style as Terminology.

**Three things that look like bugs and are not:**

1. **The response repeats values with different types** — `orgid`/`year`/`month` are numeric on the root object, `nhsnorgid`/`year`/`month` are strings inside `plans`. That is how the real API behaves.
2. **Timestamps are not RFC 3339** — `2023-09-09 11:12:12.59`, so they are typed as strings and formatted by hand. Binding them as dates would emit ISO 8601, which is well formed and not what a consumer will have to parse.
3. **The response delay never reaches `/mock` or `/health`** — a delayed `/mock` would mean turning a five-minute delay off takes five minutes, and a delayed `/health` would get the container restarted mid-test.

Normalising any of these would let a consumer write code that passes here and fails on first contact with the real endpoint, which is the one failure this service exists to prevent.

### 🧪 Testing Performed

**Automated** — full solution builds; the full `ServiceTests` suite passes (**2009 passed, 1 skipped, 0 failed**), of which **318 cover this service**.

**Against the local docker-compose stack**, verified by hand:

- **Migration** applies, rolls back and re-applies against SQL Server. The shipped index is unique with no filter and `ReportingMonth` is nullable.
- **The unfiltered index works**: a duplicate annual row is rejected with error 2601, while the same measure under another component is accepted.
- **Response shape** is a field-for-field match with the ADR's example — root values numeric, `plans` values strings, timestamps in the real format, `reporting` only ever `Y`, and `psDMRptPlanID` stable across identical requests. A non-numeric facility returns a null `orgid` with the string preserved in `plans[].nhsnorgid`.
- **Filters**: every combination of `nhsnorgid`/`name`/`year`/`month`, including `nhsnorgid` alone returning the whole plan, and malformed periods returning 400 rather than an empty plan.
- **Component isolation** in both directions, and an annual response omitting the month.
- **The delay**: a 2s delay held `/msc` for 2.03s while `/mock/delay`, `/mock/search` and `/health` all answered inside 17ms.
- **Authentication under TEST's configuration** (anonymous access off): `/mock` answers 401 unauthenticated, `/health` and `/api/mock-dmrp/info` stay open, and the contract endpoints keep working with a third-party token — the two schemes do not interfere.
- **Trimming, paging and the 404 rule**: padded values collapse onto their trimmed twins, out-of-range paging is rejected, an unknown facility is a 404 naming it while search on the same facility is a 204.

**Postman** — the `Mock DMRP API` collection in the BOTW workspace runs top to bottom against a live instance, with assertions on the response quirks, component isolation, the cadence rule, absence semantics, token errors and delay scoping.

### 🧑‍🔬 Unit Testing

- Coverage: 70.3%

Worth calling out, because each pins something that fails silently otherwise:

- `GeneratedControllerBindingTests` — pins how the NSwag base behaves once overridden. Binding sources, `[BindRequired]` and routes **do** inherit; default values and nullability **do not**, so an optional filter left as generated becomes mandatory and returns 400 for requests the contract says are valid.
- `ProblemDetailsShapeTests` — every problem carries status, title, type, a non-empty detail and a traceId; the token endpoint keeps its OAuth shape so a future tidy-up cannot quietly convert it.
- `ResponseDelayPipelineTests` — proves the delay reaches the contract endpoints and never `/mock` or `/health`.
- `ReportingPlanSemanticsTests` — component isolation and absence semantics against a real database.

Two notes on test technique. The delay's unit tests drive a `FakeTimeProvider` so a five-minute delay finishes instantly; its pipeline tests use the real clock, because a fake one races through `TestServer` — assertions there use a lower bound for a configured delay and a five-minute delay for the never-delayed paths, so a loaded agent cannot flip either. Ran repeatedly under the full parallel suite to confirm.

New test-only dependency: `Microsoft.Extensions.TimeProvider.Testing`.

### 📓 Documentation Updated

- **`DotNet/MockDmrpApi/README.md`** (~730 lines) — the two surfaces, the contract-first workflow and how to replace the spec, the query parameters, the response quirks and why they must not be tidied up, the cadence rule, the filtered-index hazard, the availability switch, the two authentication systems, error responses, and an open-questions list separating what the ADR settled from what is still guessed.
- **`app-config.yaml`** — service entries; the Link authentication keys are already global, so nothing service-specific was added for them.
- **`docs/config-key-inventory.md`** — regenerated.
- **TestRail** — the 37 existing cases were audited against this branch. All seven conflicts between the cases and the service now have rulings; three changed the service and are in this PR, four are edits QA will apply to the cases.
- **Jira** — deviations from the ADR's proposed mock design are recorded on the story, as the ADR asks.

## Summary by CodeRabbit

* **New Features**
  * Added a mock DMRP API supporting monthly medicine and annual patient-safety reporting plans.
  * Added authenticated reporting-plan retrieval, search, creation, updates, and deletion.
  * Added mock token issuance, configurable response delays, health checks, and structured error responses.
  * Added Docker Compose, deployment, configuration, and CI/CD support.
* **Documentation**
  * Added API contract and setup, configuration, authentication, and usage documentation.
* **Tests**
  * Added comprehensive unit, integration, persistence, authentication, validation, and response-behavior coverage.


DotNet/MockDmrpApi/Presentation/Controllers/MockController.cs
```diff
@@ -0,0 +1,451 @@
...
+        var sanitizedFacilityId = facilityId.SanitizeAndRemove();
+        if (string.IsNullOrWhiteSpace(sanitizedFacilityId))
+        {
+            return MissingFacilityId();
+        }
+
+        var (records, metadata) = await _reportingPlans.GetByFacilityAsync(
+            sanitizedFacilityId, pageSize ?? 10, pageNumber ?? 1, cancellationToken);
+
+        if (records.Count == 0)
```
smailliwcs: Should check `metadata.TotalCount` rather than `records.Count`?

DotNet/MockDmrpApi/Application/Services/AuthTokenService.cs
```diff
@@ -0,0 +1,154 @@
...
+    public AuthTokenService(IOptions<DmrpApiSettings> settings)
+    {
+        ArgumentNullException.ThrowIfNull(settings);
+        _settings = settings.Value;
+
+        var keyBytes = Encoding.UTF8.GetBytes(_settings.SigningKey ?? string.Empty);
+        if (keyBytes.Length < MinimumSigningKeyBytes)
+        {
+            // Fail at startup rather than on the first token request, where it would
+            // surface as an opaque 500 from an endpoint that looks unrelated.
```
smailliwcs: Since `AuthTokenService` is registered as a lazily constructed singleton, won't the execution of its constructor be deferred until it's first used as a dependency?  I.e., does an invalid signing key actually cause startup failure as suggested in the comment?

DotNet/MockDmrpApi/Program.cs
```diff
@@ -0,0 +1,151 @@
...
+using LantanaGroup.Link.Shared.Application.Extensions;
+using LantanaGroup.Link.Shared.Application.Extensions.Security;
+using LantanaGroup.Link.Shared.Application.Middleware;
+using LantanaGroup.Link.Shared.Domain.Repositories.Implementations;
+using LantanaGroup.Link.Shared.Domain.Repositories.Interceptors;
+using LantanaGroup.Link.Shared.Domain.Repositories.Interfaces;
+using Microsoft.AspNetCore.Diagnostics.HealthChecks;
+using System.Reflection;
+
+var builder = WebApplication.CreateBuilder(args);
```
smailliwcs, line 17: We're missing the `builder.Configuration.AddStandardEnvironmentConfiguration` call that is included just after `CreateBuilder` in all other .NET services.

DotNet/MockDmrpApi/Presentation/Controllers/MockController.cs
```diff
@@ -0,0 +1,451 @@
...
+        CancellationToken cancellationToken = default)
+    {
+        var criteria = new ReportingPlanSearchCriteria
+        {
+            FacilityId = Blank(facilityId),
+            Component = Blank(component),
+            Measure = Blank(measure),
+            ReportingMonth = reportingMonth,
+            ReportingYear = reportingYear,
+            IsReporting = Blank(isReporting),
```
smailliwcs, line 131: Need to sanitize `IsReporting` the same way in `Create`/`Update` as well?  Otherwise you could persist an unsanitized value that becomes unsearchable (because it changes during sanitization).

Scripts/set_kubernetes_services.bat
```diff
@@ -30,4 +30,5 @@ kubectl set -n %namespace% image -n %namespace% deployment report-deploy report=
 kubectl set -n %namespace% image -n %namespace% deployment submission-deploy submission=%container-registry%/link-submission:%image%
 kubectl set -n %namespace% image -n %namespace% deployment tenant-deploy tenant=%container-registry%/link-tenant:%image%
 kubectl set -n %namespace% image -n %namespace% deployment terminology-deploy terminology=%container-registry%/link-terminology:%image%
-kubectl set -n %namespace% image -n %namespace% deployment validation-deploy validation=%container-registry%/link-validation:%image%
\ No newline at end of file
+kubectl set -n %namespace% image -n %namespace% deployment validation-deploy validation=%container-registry%/link-validation:%image%
+kubectl set -n %namespace% image -n %namespace% deployment mock-dmrp-deploy mock-dmrp=%container-registry%/link-mock-dmrp:%image%
```
smailliwcs, line 34: Need a similar addition in `build_and_push_and_set.py`?

smailliwcs (approved):

DotNet/MockDmrpApi/Presentation/Controllers/MockController.cs
MikeAtPinnacle: 😕 If `metadata.TotalCount` was zero, wouldn't `records.Count` be zero?

DotNet/MockDmrpApi/Application/Services/AuthTokenService.cs
MikeAtPinnacle: Valid point, I will adjust

DotNet/MockDmrpApi/Presentation/Controllers/MockController.cs
MikeAtPinnacle: fixed

DotNet/MockDmrpApi/Program.cs
MikeAtPinnacle: added

Scripts/set_kubernetes_services.bat
MikeAtPinnacle: @smailliwcs / @michael-misiaszek  Do you know what is actually used?

I will update the .py file, but if the .bat file isn't used, we should probably remove it from the repo.

DotNet/MockDmrpApi/Presentation/Controllers/MockController.cs
smailliwcs: If this is a paginated query, doesn't `records` represent the current page?  Which could be empty, even if `metadata.TotalCount` is nonzero, if you paged past the end of the result set.

Scripts/set_kubernetes_services.bat
smailliwcs: I assume _some_ of these scripts are used in our ADO pipelines or other workflows, but I'm not sure which.

DotNet/MockDmrpApi/Presentation/Controllers/MockController.cs
MikeAtPinnacle: I am corrected this

PR #1848, opened 2026-08-19
LEGLINK-828: Tenant service integration with the DMRP module

MikeAtPinnacle:
### 🛠️ Description of Changes

The Tenant facility endpoints now resolve their state-changing operations through an interface the DMRP module replaces with its own implementation when `DMRP:Enabled` is set, so a facility's scheduled reports are derived from its DMRP reporting plans instead of being chosen by the caller.

**No BFF changes, and no new routes.** The ADR's *BFF Reverse Proxy and Shared Interface* section was written when DMRP was expected to be a separately deployed service. It is a class library hosted in-process by Tenant, and `api/dmrp/**` already proxies to the Tenant cluster, so the reverse-proxy hop it describes has no counterpart in the code. The facility URL surface is identical with the flag on or off; only the behavior behind POST, PUT and DELETE differs.

**Backend**

- `IFacilityOperations` (DMRP) declares the five state-changing facility operations; `TenantFacilityOperations` (Tenant) is the host's implementation, holding the manager call and the `ScheduleService` reconciliation lifted out of the controller.
- `DmrpFacilityOperations` wraps it when the flag is on: create and update refuse a caller-supplied `scheduledReports` and derive one from the facility's reporting plans; hard delete removes the facility first and its plans second; soft delete and restore pass through.
- `DbBackedReportingPlanSource` reads `FacilityReportingPlans` joined to `MeasureMappings` for the period the facility is currently in, read in its own timezone. `IReportingPlanSource` is the seam the DMRP API client slots into with LEGLINK-698/701.
- `AddDmrpModule` gains a type parameter naming the host's implementation, so a host that supplies none fails to compile rather than at startup.
- `FacilityController` calls the interface; its validation, sanitization and status codes are unchanged.

**Admin UI**

- The facility form required at least one scheduled report, which the API now refuses — no input satisfied both, making facility create and edit impossible with DMRP on. `ScheduledReportsValidator` takes the flag and drops that rule; the report pickers are replaced by a line saying the schedule is derived; submit sends empty arrays.

**Two defects fixed in the measure mappings endpoints**

Both were found while documenting these endpoints, and predate this work:

- Deleting a measure mapping that reporting plans still reference answered `404`. The row plainly exists and is readable, so reporting it missing said the opposite of what happened. It now answers `409` with a message naming the reason, checked before the delete is attempted so a refusal changes nothing. Delete-all is refused the same way while any plan exists. A mapping that genuinely does not exist still answers `404`.
- `DELETE /api/dmrp/measure-mappings/{id}` was the only write endpoint on the controller without `[ValidateAntiForgeryOrBearerToken]`.

**Configuration**

- `DMRP:Enabled` moves from the Tenant section of `app-config.yaml` to `global`. It is one decision for the deployment, and unlabeled means every service reads the same answer. The Angular app cannot read App Configuration, so the same decision reaches it via `LINK_DMRP_ENABLED` on its container. The two must agree — the UI switched on ahead of the services quietly creates facilities that report nothing, while the reverse fails loudly.
- No migration (no entity changes) and no new config key. `DMRP:Enabled` is absent from every environment store and defaults to `false`, so DMRP is currently off everywhere.

**Deviations from the ADR, for QA**

1. No BFF changes and no `api/dmrp/facility/**` routes, as above.
2. `scheduledReports` must be sent as **empty arrays**, not omitted — its arrays are non-nullable, so an absent block fails model binding before the API sees it.
3. Reporting plans are read from the table, **not fetched from the DMRP API**. That arrives with LEGLINK-698/701, so until then a facility gets an empty schedule unless its plan rows are seeded first.

### 🧪 Testing Performed

Full .NET suite green: **2280 passed, 1 skipped, 0 failed**. `link-cloud.sln` builds with 0 errors. `ng build` clean; `validate_app_config_schema.py` passes and `check_required_config.py` reports every required key present in dev, qa and test.

End-to-end against the local docker-compose stack, Tenant with `DMRP:Enabled=true`, seeding two measure mappings (HOB to monthly, HTCDI to daily) and reporting plans for the current period:

| Step | Result |
|---|---|
| Module loads with the flag on | `DMRP module enabled: True` |
| `POST /api/Facility` **with** `scheduledReports` | `400` naming DMRP as the source |
| `POST /api/Facility` with **empty** `scheduledReports` | `201`, daily = RPS dQM, monthly = ACH dQM, weekly empty |
| `PUT` with a caller-supplied schedule | `400` |
| `PUT` with empty schedule | `200`, schedule re-derived |
| `softDelete` then re-query plans | 2 rows kept |
| `restore` then re-query plans | 2 rows kept |
| hard `DELETE` then re-query plans | 0 rows; log confirms `Deleted 2 reporting plan(s)` |
| Flag off: `api/dmrp/**` | `404` |
| Flag off: caller's `scheduledReports` | `201`, honored verbatim |
| Flag off: full facility lifecycle | all `204` |

That run found and fixed a real defect: the refusal message originally said "resubmit without scheduledReports", which model binding makes impossible. It now names the empty-arrays remedy, verified end to end.

The measure mappings delete fix was verified the same way:

| Request | Result |
|---|---|
| `DELETE /{id}` on a referenced mapping | `409` and "referenced by one or more facility reporting plans" |
| `GET /{id}` afterwards | `200` — the refusal changed nothing |
| `DELETE` all while a plan exists | `409` |
| `DELETE /{id}` after clearing the plan | `204`, then `404` on re-read |
| `DELETE /{id}` on a mapping that never existed | `404` — still distinct |

Measure-definition validation was disabled for the facility run (`MeasureConfig__CheckIfMeasureExists=false`) via an override file outside the repo, so `docker-compose.yml` carries no test-only change. **Not yet covered:** a seeded full-stack run confirming realistic dQM mappings produce a facility Tenant accepts with that validation on.

### 🧑‍🔬 Unit Testing

- Coverage: 57.5%
  - `DmrpFacilityOperationsTests` — schedule derivation, distinct dQMs when two measures map to one, unmapped measures excluded, reporting period read in the facility's timezone with a UTC fallback for an unusable one, refusal on create and update, an explicitly empty block accepted, delete ordering (facility first, plans second) and plans kept when the inner delete throws.
  - `DbBackedReportingPlanSourceTests` — SQLite-backed filtering by facility and period, `IsReporting=false` excluded, unmapped measure returned with no dQM.
  - `DmrpModuleExtensionsTests` — the module fronts the host's operations when enabled and leaves them alone when disabled.
  - `DmrpFacilityOperationsIntegrationTests` — the wiring end to end: the module really does front the host, and the schedule comes from rows read out of the database.
  - `MeasureMappingsControllerTests` — delete refused with `409` while a reporting plan references the mapping (and the mapping still readable afterwards), `204` when nothing references it, `404` when it does not exist.
  - `FacilityControllerTests.Facility_lifecycle_keeps_the_scheduled_jobs_in_step` — asserts Quartz job presence across create, soft delete, restore and delete. Added because the existing controller tests assert status codes only, and job re-creation was dropped from restore during this work without any test noticing.
  - `ScheduledReportsValidator.spec.ts` — 5 specs covering both flag states and the parameter default.

### 📓 Documentation Updated

- XML documentation on all 15 DMRP controller endpoints (params, response codes, validation rules), which Swagger picks up through the DMRP XML file Tenant already feeds it. The same text is on the matching requests in the **Link Admin BFF** Postman collection, including the new `409` behavior.
- `app-config.yaml` — `DMRP:Enabled` re-catalogued as a global system switch, with the UI/services agreement requirement and its asymmetric failure modes recorded.
- The DMRP UI flag is temporary and expected to end up permanently on, so it is written to be deleted: every site is marked `DMRP feature flag` and the removal recipe lives on `AppConfig.dmrpEnabled`.

## Summary by CodeRabbit

* **New Features**
  * Added DMRP-based facility scheduling, deriving daily, weekly, and monthly reports from reporting plans.
  * Added a system-wide DMRP feature switch for enabling or disabling plan-driven scheduling.
  * Facility creation, updates, deletion, soft deletion, and restoration now keep schedules and reporting plans synchronized.
* **Bug Fixes**
  * Prevented deletion of measure mappings still referenced by reporting plans, returning a conflict response.
  * Unmapped measures are excluded from generated schedules.
* **UI**
  * Facility configuration now explains when schedules are controlled by reporting plans and adjusts validation accordingly.
* **Documentation**
  * Expanded endpoint documentation and error-response details.


DotNet/DMRP/Business/DmrpFacilityOperations.cs
```diff
@@ -0,0 +1,268 @@
...
+        /// until the delete is retried or the service restarts - <c>ScheduleService.StartAsync</c>
+        /// rebuilds jobs for every facility that is not deleted, and removing them is idempotent. That
+        /// heals; a stranded reporting plan does not.
+        /// </para>
+        /// </remarks>
+        public async Task DeleteAsync(string facilityId, CancellationToken cancellationToken = default)
+        {
+            await _reportingPlanRepository.StartTransactionAsync(cancellationToken);
+
+            int removed;
```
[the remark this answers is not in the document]
MikeAtPinnacle: Not taking this one — the mechanism described doesn't hold.

`SanitizeForLog` is declared `public static string SanitizeForLog(this object? obj)` and returns null for a null input (`DotNet/Shared/Application/Services/Security/StringSecurityExtensions.cs:33-35`), so a null `facilityId` logs as null rather than throwing an NRE at line 116.

The dangling-transaction concern doesn't apply either: anything the host's `DeleteAsync` throws is caught at line 109 and routed through `RollbackQuietlyAsync` before being rethrown, so the transaction is closed on every path out.

Null is also unreachable here. The only caller is `FacilityController.DeleteFacility`, which returns 404 at line 344 when the facility lookup misses, and the `{facilityId}` route segment can't bind null.

DotNet/DMRP/Controllers/MeasureMappingsController.cs
```diff
@@ -206,6 +302,10 @@ public async Task<IActionResult> DeleteMeasureMapping(string id, CancellationTok
             {
                 await _manager.DeleteAsync(id, cancellationToken);
             }
+            catch (MeasureMappingInUseException ex)
+            {
+                return Problem(ex.Message, statusCode: StatusCodes.Status409Conflict, title: "Conflict");
+            }
```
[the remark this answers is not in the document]
MikeAtPinnacle: Out of scope for this PR as it stood, and it wouldn't have added anything at the time.

The PR touched no LinkSdk file when this was written (`git diff origin/dev...HEAD -- DotNet/LinkSdk` was empty), and the bulk delete endpoint already existed on `dev`, so the gap predates the change. More to the point, `IDmrpServiceClient` had no consumers anywhere in the repo — the analogous `DeleteFacilityReportingPlansAsync` is declared and never called — so a new method would not have been exercised by anything.

That has since changed in a way worth flagging: a later commit on this branch (8c4b2414b) gives the client its first consumer, in the automation facility setup. Doing so surfaced a real defect in the same file — `SearchMeasureMappingsAsync` was reading the measure-mapping collection route, which only accepts POST, so the call answered 404. That's indistinguishable from the DMRP module being switched off, and nothing noticed because there were no callers. It now reads `/dmrp/measure-mappings/search` and passes the measure, dQM and frequency filters the API already supports.

Still not adding the bulk delete, since nothing calls it.

DotNet/ServiceTests/UnitTests/DMRP/DmrpFacilityOperationsTests.cs
```diff
@@ -0,0 +1,373 @@
...
+                new ReportingPlanEntry("NEWMEASURE", string.Empty, Frequency.Adhoc));
+
+            var facility = Facility();
+
+            await CreateOperations().CreateAsync(facility);
+
+            Assert.Equal(new[] { "dqm-monthly" }, facility.ScheduledReports.Monthly);
+            Assert.Empty(facility.ScheduledReports.Daily);
+            Assert.Empty(facility.ScheduledReports.Weekly);
+        }
```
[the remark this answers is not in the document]
MikeAtPinnacle: Taken — fixed in 503bab0b6.

You're right that the assertion wasn't load-bearing. `DqmsFor` is only invoked for Daily, Weekly and Monthly, and `Frequency.Adhoc` matches none of them, so the `NEWMEASURE` entry was excluded by the frequency mismatch before `!string.IsNullOrWhiteSpace(e.DQM)` was ever consulted.

Changed that entry to `Frequency.Monthly` so the missing dQM is the only thing that can exclude it, and confirmed by mutation: deleting the dQM filter from `DqmsFor` now fails this test, where previously it passed.

Also split the adhoc case into a test of its own rather than folding both exclusions into one assertion — the file had no coverage of the adhoc frequency at all, so that's net new.

MikeAtPinnacle:
Dispositions for the remaining review-body items, so nothing is left ambiguous.

**Taken (503bab0b6)**

- **`IsStillReferenced` backstop coverage.** Added: SQLite foreign-key (787) and not-null (1299) failures both surface as `MeasureMappingInUseException`, a failure nested two levels deep still does — which is what makes the chain walk load-bearing rather than a plain `InnerException` check — and an unrelated failure (SQLITE_BUSY) does not. The SQL Server codes the same backstop recognises (547, 515) cannot be reached from a test: `SqlException` has no public constructor and is only ever produced by the driver. That's recorded in the test's doc comment so the omission reads as deliberate.
- **Registration-order hazard around `RemoveAll<IFacilityOperations>()`.** Guarded. The check accepts *either* the interface or the host's implementation type being registered, rather than only the interface — `DmrpIntegrationTestFixture` legitimately supplies the host by concrete type, and a narrower check would have broken it while still not catching anything extra.
- **The intermediate `unmapped` list.** Removed.

**Not taking**

- **Redundant facility read in `TenantFacilityOperations.SoftDeleteAsync`.** The duplicate read is real, but threading the already-loaded facility through `IFacilityOperations` would couple the seam to the order the controller happens to do things in, to save one read on a delete. Not a good trade.

**Backlog rather than blockers**

- **Moving `IFacilityOperations` out of the DMRP assembly.** Valid direction and consistent with `DotNet/Shared` being the contract layer, but `Tenant.csproj` already carried a `ProjectReference` to `DMRP.csproj` on `dev` — this PR widened an existing inversion rather than introducing one.
- **A real-host rollback integration test.** Checked the assumption it would be guarding: `FacilityRepository` and the module's `EntityRepository<FacilityReportingPlan, TenantDbContext>` are both scoped over the same `TenantDbContext`, so one transaction genuinely does span the host's write and the module's. The design holds by construction; a test would guard a future regression, such as a move to `IDbContextFactory`, rather than a present gap. Noting it here so the reasoning survives.

arianamihailescu (commented):
The description says: 
"No BFF changes, and no new routes. The ADR's BFF Reverse Proxy and Shared Interface section was written when DMRP was expected to be a separately deployed service."

But in the ADR I read : 
"A new DMRP C# project should be created in the Link Cloud codebase. This is intended to be a project that enables additional functionality on the Tenant service and is not going to be a separately deployed service."

They seem to be contradictory.

MikeAtPinnacle:
> The description says: "No BFF changes, and no new routes. The ADR's BFF Reverse Proxy and Shared Interface section was written when DMRP was expected to be a separately deployed service."
> 
> But in the ADR I read : "A new DMRP C# project should be created in the Link Cloud codebase. This is intended to be a project that enables additional functionality on the Tenant service and is not going to be a separately deployed service."
> 
> They seem to be contradictory.

@arianamihailescu 
DMRP is a new project, and Tenant uses DMRP implementations when DMRP is active.  New routes support DMRP-specific items, but the existing routes should work as before.  I did have to make a change so that adding a Tenant wouldn't fail with DMRP enabled.  I suspect [LEGLINK-707](https://lantana.atlassian.net/browse/LEGLINK-707) will change this.

[LEGLINK-707]: https://lantana.atlassian.net/browse/LEGLINK-707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Web/Admin.UI/src/app/components/validators/ScheduledReportsValidator.ts
```diff
@@ -15,8 +24,9 @@ export function ScheduledReportsValidator(): ValidatorFn {
 
     const errors: ValidationErrors = {};
 
-    // Check if at least one report is entered
-    if (allReports.length === 0) {
+    // Check if at least one report is entered. DMRP feature flag: this whole check goes when the
+    // flag is retired, because a derived schedule is never chosen on this form.
+    if (!dmrpEnabled && allReports.length === 0) {
```
arianamihailescu: We do not need to validate if scheduledReports are added. They are not required on the server side so we do not need to do that in UI.
Does it mean that we do not need dmrpEnabled flag in UI?

DotNet/DMRP/Business/DmrpFacilityOperations.cs
```diff
@@ -0,0 +1,266 @@
...
+            IFacilityOperations hostImplementation,
+            IReportingPlanSource reportingPlans,
+            IFacilityReportingPlanManager reportingPlanManager,
+            IEntityRepository<FacilityReportingPlan> reportingPlanRepository,
+            TimeProvider timeProvider)
+        {
+            _logger = logger ?? throw new ArgumentNullException(nameof(logger));
+            _hostImplementation = hostImplementation ?? throw new ArgumentNullException(nameof(hostImplementation));
+            _reportingPlans = reportingPlans ?? throw new ArgumentNullException(nameof(reportingPlans));
+            _reportingPlanManager = reportingPlanManager ?? throw new ArgumentNullException(nameof(reportingPlanManager));
```
arianamihailescu: The DbBackedReportingPlanSource is reading from reporting plan table (the reading part is there) but I do not see the real implementation for the ReportingPlanSource that will go call  the MOCK API to get the  data and write to the table ( like ADR says).

Web/Admin.UI/src/app/components/validators/ScheduledReportsValidator.ts
MikeAtPinnacle: This was done because the existing code reported an error (line 30) if no reports were entered.

DotNet/DMRP/Business/DmrpFacilityOperations.cs
MikeAtPinnacle: That is in [LEGLINK-701](https://lantana.atlassian.net/browse/LEGLINK-701) - which is actually assigned to you in the next iteration.

Web/Admin.UI/src/app/components/validators/ScheduledReportsValidator.ts
MikeAtPinnacle: As discussed, I removed this check

arianamihailescu (approved):
Looks good

PR #1858, opened 2026-08-25
LEGLINK-1068: Align ACH measure bundles and predict ABS from measure CQL

nvmLantana:
### 🛠️ Description of Changes

LEGLINK-1068. Align Automation ACH Daily and Monthly measure bundles with Validation 2.0.0-cibuild, and derive ABS instance-level CQL prediction from those embedded bundles instead of frozen family profiles.

Measure bundles: Validation Measure + primary Library, FHIRHelpers 4.0.2 and NHSNHelpers 0.0.002 kept so CQF can resolve includes, Daily Discharge Disposition ValueSet URL aligned with CQL (`clinical-discharge-disposition`).

ABS prediction: `CqlInstanceFilterAnalyzer` / `CqlMeasureBundleModel` parse reachable SDE defines from the measure-bundle CQL. DiagnosticReport Others (empty-category / not-category), Location.partOf follow during acquisition, Monthly SDE Location limited to IP encounter locations, DiagnosticReport dates match `effective[x]` only, default Observation category query includes `survey`, and Daily Observations on the last second of PeriodEnd are predicted. Lab Observations are stamped with point `effectiveDateTime` (Thetis shape) so they fall inside a one-day Daily window.

Adds `AdhocReportDailyAchTest` and unit coverage for the CQL simulator, acquisition simulator, and embedded ACH bundles. No Measure Eval Java changes.

### 🧪 Testing Performed

- Unit tests: `CqlFilterSimulatorTests`, `EmbeddedAchMeasureBundleTests`, `QueryPlanAcquisitionSimulatorTests`, `FhirBundleGeneratorTests`, `OrgResourceMapPredictionFilterTests`, `ScenarioSeedServiceTests`.
- Backend `AdhocReportDailyAchTest` added for Daily ACH.
- Silo Daily adhoc on the first measure-only swap still scored IP=0 until the discharge ValueSet was aligned. Silo Monthly adhoc `a84718ed` on the aligned Monthly package scored `initial-population=[1]` / `REPORTABLE=[true]`.

### 🧑‍🔬 Unit Testing

- Coverage: CqlFilterSimulatorTests, EmbeddedAchMeasureBundleTests, QueryPlanAcquisitionSimulatorTests, FhirBundleGeneratorTests, OrgResourceMapPredictionFilterTests, ScenarioSeedServiceTests

### 📓 Documentation Updated

Automation README CQL filter simulator section (prediction now reads embedded measure-bundle CQL). Short notes in Automation.Link, Automation.UI, and BackendE2ETests READMEs.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -178,6 +178,7 @@ jobs:
           # name|category|trx-file
           TESTS=(
             "Adhoc Report|AdhocReportTest|adhoc-report-test-results.trx"
+            "Adhoc Report Daily ACH|AdhocReportDailyAchTest|adhoc-report-daily-ach-test-results.trx"
             "API Stability|ApiStabilityTest|api-stability-test-results.trx"
             "Automation UI API Smoke|AutomationUiSmokeTest|automation-ui-api-smoke-test-results.trx"
             "Report Scheduled|ReportScheduledTest|report-scheduled-test-results.trx"
--- a/DotNet/Automation.Link/README.md
+++ b/DotNet/Automation.Link/README.md
@@ -150,12 +150,9 @@ There is no tolerant `actual >= expected` mode. Pipeline-derived types (`Patient
 strict mode holds.
 
 Prediction is intentionally narrower than "generated and acquired". The manifest also applies
-CQL type reachability and resource-level SDE filtering from `Automation.CqlFilterSimulator`.
-This includes patient-context retrieval checks for referenced resources. For example, an
-acquired `Specimen` is only expected in ABS when the selected measure's SDE logic would return
-it for the evaluated patient: ACH Monthly requires patient-owned specimen collection overlap
-with IP, ACH Daily requires a qualifying respiratory-pathogen lab observation reference, and
-Hypoglycemic requires patient-owned collection fully during IP.
+CQL type reachability and resource-level SDE filtering from `Automation.CqlFilterSimulator`,
+which derives instance filters from the selected measure's embedded bundle CQL (not from a
+frozen measure-family profile).
 
 ### 5.2 Validators
 
--- a/DotNet/Automation.UI/README.md
+++ b/DotNet/Automation.UI/README.md
@@ -611,7 +611,7 @@ Why SSE here:
      - Data Acquisition org-location configuration for the run facility, and
      - generation/prediction modeling so expected ABS counts are org-scope aware.
    - `QueryPlanAcquisitionSimulator` runs per-patient with the resolved clinical period.
-   - `CqlFilterSimulator` runs per-patient over the patient's qualifying measures only.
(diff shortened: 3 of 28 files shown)


smailliwcs (approved):
Description has this:

> After LEGLINK-186 (async `@KafkaListener` consume of `ResourcesNormalized`), that first read can happen before Normalization finishes `UpdateCorrelationCacheAsync`.

But if I'm reading the code correctly, that shouldn't be possible, because Normalization doesn't produce `ResourcesNormalized` until after all `UpdateCorrelationCacheAsync` calls are awaited:

https://github.com/lantanagroup/link-cloud/blob/2412d21ed9ec3909b07c3bbf2d5e203b5d220111/DotNet/Normalization/Listeners/ResourcesAcquiredListener.cs#L318-L321

Or is there somehow a delay between `UpdateCorrelationCacheAsync` completing and the key becoming available to readers in Redis?  In which case, what the heck?  Regardless, I'm approving, because this obviously fixes *something*, but would be nice to know what that is.

nvmLantana:
> Description has this:
> 
> > After LEGLINK-186 (async `@KafkaListener` consume of `ResourcesNormalized`), that first read can happen before Normalization finishes `UpdateCorrelationCacheAsync`.
> 
> But if I'm reading the code correctly, that shouldn't be possible, because Normalization doesn't produce `ResourcesNormalized` until after all `UpdateCorrelationCacheAsync` calls are awaited:
> 
> https://github.com/lantanagroup/link-cloud/blob/2412d21ed9ec3909b07c3bbf2d5e203b5d220111/DotNet/Normalization/Listeners/ResourcesAcquiredListener.cs#L318-L321
> 
> Or is there somehow a delay between `UpdateCorrelationCacheAsync` completing and the key becoming available to readers in Redis? In which case, what the heck? Regardless, I'm approving, because this obviously fixes _something_, but would be nice to know what that is.

A very good point. I am also chasing down a possibly related and in actuality root cause of this issue. Doing testing now. Will reshape the story and related ticket with findings and ask for review again, if so.

nvmLantana:
You're right, that race isn't possible. Normalization awaits `UpdateCorrelationCacheAsync` before producing `ResourcesNormalized`, so Measure Eval cannot read the correlation hash before the write completes.

The empty-cache retry is reverted. What was actually aborting Daily is a CQF 4.5.1 NPE in `evaluateSdes`: `NHSNAcuteCareHospitalDailyInitialPopulation.supplementalData` names five identifiers CQL never defines (`SDE ACH Daily Observation` / Specimen / DiagnosticReport / DiagnosticReport Result from Lab / ServiceRequest; the library still has the `SDE RPS*` names). That runs before Initial Population, the handler swallows the NPE, and we still publish `isReportable=false`.

This PR now drops unmatched `supplementalData` at compile time. Silo Daily after the rebuild: 0 `expressionResult` NPEs, those five names logged as dropped, 10/10 `REPORTABLE=[true]` (was all false).

PR #1859, opened 2026-08-25
TECH_DEBT: Pin db-schema-creation to a release and check it against the target environment

MikeAtPinnacle:
### 🛠️ Description of Changes

The DB schema pipeline generated its EF script from whatever branch the run was queued on, and its two release parameters only fed an advisory change report. Nothing tied the generated schema to the code an environment was actually running. This PR makes `current_release` determine the output, replaces the hand-typed baseline with the commit deployed to the selected environment, and refuses to apply when the two are incompatible.

**Why it matters:** a run queued on `dev` pushed tip-of-dev schema into an environment deployed from an older release. That is how QA2 had `Facilities.Vendor` dropped while the Tenant image still selected the column, which surfaced in the Admin UI as an unrelated "No Acquisition Logs Found".

- **`current_release` is now honored.** Every schema-generation job gets its own `checkout: self` with `fetchDepth: 0` and `fetchTags: true`, then `git checkout --detach "$(current_release)"` before running `dotnet-ef`. The generated script comes from the working tree, so without this the parameter was a label with no effect on the output. Each job echoes the commit it generated from.
- **`previous_release` is replaced by an `environment` selector** (`scale-test` | `scale-qa` | `scale-qa2`). The diff baseline now comes from `Scripts/get_deployed_commit.py`, which reads `<BASE_URL>/api/info` for the chosen environment and sets `FromCommit`. Adds the `link-cloud-variables` variable group, which holds the `*_BASE_URL` values that script reads.
- **New guard comparing `current_release` to the deployed commit:**
  - same commit → passes
  - schema **ahead** of deployed code → warning, and it lists every migration file in the gap so the reviewer can judge whether any of them are destructive before approving the apply
  - schema **behind** deployed code → fails (the running services expect migrations the script does not contain)
  - **no shared ancestry** → fails (usually means the run was queued on the wrong branch)
- **Apply stages are now gated.** Each stage gets `dependsOn: Build` and `condition: eq('${{ parameters.environment }}', '<env>')`, replacing `condition: always()`. Two consequences: a run targets exactly one environment instead of all three, and a failed comparison in `Build` blocks the apply instead of being advisory.
- **Change report fixes.** The service-to-migration-path map is now explicit and covers `Submission`, whose migrations live in `DotNet/Submission.Data/Migrations` and were silently never reported under the old `DotNet/<service>/migrations` pattern. Three near-identical loops (DotNet / DataAcquisition / Java) fold into one.
- Fixes the `db.aliase` typo in the QA2 apply stage's job display name.

**How to use it:** queue the pipeline, pick the target `environment`, and set `current_release` to the tag or SHA that environment is deployed from (or is about to be deployed from). The `Build` stage reports which services have migration changes between the deployed commit and `current_release`, and either passes, warns with the list of migrations in the gap, or fails. Then manually trigger the one apply stage matching the chosen environment. The old workflow — typing a previous-release tag from memory and having all three apply stages available regardless — is gone.

### 🧪 Testing Performed

Static verification against the repo, since the pipeline itself only runs in Azure DevOps:

- The edited YAML parses and still resolves to the four expected stages (`Build`, `Apply_TEST_Update_SQL_DBs`, `Apply_QA_Update_SQL_DBs`, `Apply_QA2_Update_SQL_DBs`).
- All eleven paths in the new migration-path map exist in the repo, including the two that do not follow the plain `DotNet/<service>/Migrations` shape (`DotNet/DataAcquisition.Domain/Migrations`, `DotNet/Submission.Data/Migrations`). Confirmed the old pattern could not have matched Submission.
- Confirmed the variable contract with `Scripts/get_deployed_commit.py`: it emits `##vso[task.setvariable variable=FromCommit]`, which is job-scoped and not an output variable, and both consumers (the comparison step and the change-report step) are steps in the same `getChangedFiles` job, so `$(FromCommit)` resolves.
- Walked the comparison logic against the four cases it distinguishes (equal / ancestor / descendant / unrelated) using `git merge-base --is-ancestor` semantics.

Not yet validated by a queued pipeline run — the `link-cloud-variables` group and the environment base URLs need to resolve in Azure DevOps, and the `--detach` checkout in each generation job should be confirmed against a real tag before this is relied on for a release.

### 🧑‍🔬 Unit Testing

- Coverage: 0.0%

### 📓 Documentation Updated

No documentation changes. The pipeline's parameters are self-describing via their `displayName` values, and the rationale for each change is captured in comments in the YAML itself.

### DIFF
--- a/Azure_Pipelines/azure-pipelines.db-schema-creation.yaml
+++ b/Azure_Pipelines/azure-pipelines.db-schema-creation.yaml
@@ -9,11 +9,16 @@ pool:
 
 variables:
 - group: 'MSSQL_DBCreation'
+- group: link-cloud-variables
 
 parameters:
-- name: previous_release
-  displayName: "What is the Tag or Commit SHA of the Previous Release? (i.e. : v0.2.0-beta.1)"
+- name: environment
+  displayName: Which environment is this for? (TEST | QA | QA2)
   type: string
+  values:
+    - scale-test
+    - scale-qa
+    - scale-qa2
 
 - name: current_release
   displayName: "What is the Tag or Commit SHA of the Current Release? (i.e. : v0.2.0-beta.2)"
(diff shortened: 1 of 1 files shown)


michael-misiaszek (approved):
Looks good Mike.

PR #1861, opened 2026-08-26
LEGLINK-1084: Chunk bundle validation, cache Terminology, and size heap from RAM

nvmLantana:
### 🛠️ Description of Changes

LEGLINK-1084. Chunk ReadyForValidation bundle validation, cache Terminology lookups, construct ValidationService at startup, and size the Validation heap from container RAM.

Bundle validation uses `bundle-validation-parallelism=4` and `bundle-validation-batch-size=32` so a large patient bundle is not held as one in-memory validate. `ValidationCacheService` / `RemoteTermServiceValidation` cache ValueSet and code membership results across resources. `ValidationService` is an eager singleton so IG packages load at startup instead of on the first patient.

Heap is `-XX:MaxRAMPercentage=75.0` in the Validation Dockerfile and compose (no fixed `-Xmx`). Local compose sets `mem_limit: 3g` (~2.25g heap) and health `start_period: 45s` so package load can finish.

### 🧪 Testing Performed

- Unit tests: `ValidationCacheServiceCachingTest`, `RemoteTermServiceValidationTest`, `ValidationServiceTest`.
- Silo ACH Monthly adhoc `a84718ed` reproduced the OOM on unpatched Validation after ReadyForValidation. This branch has not been re-proven on that silo stack in this PR description.

### 🧑‍🔬 Unit Testing

- Coverage: 59.9%

### 📓 Documentation Updated

None. Dockerfile / compose heap comments only.

### DIFF
--- a/Java/validation/src/main/java/com/lantanagroup/link/validation/configs/LinkConfig.java
+++ b/Java/validation/src/main/java/com/lantanagroup/link/validation/configs/LinkConfig.java
@@ -13,6 +13,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicInteger;
 
 @Configuration
 @ConfigurationProperties("link")
@@ -49,25 +50,18 @@ public class LinkConfig {
     private List<ValidationResultIgnoreRuleConfig> validationResultIgnoreRules = new ArrayList<>();
 
     /**
-     * How many bundle chunks validate at once. Memory is bounded by this times one
-     * in-flight HAPI validation (entries inside a chunk run sequentially).
+     * How many bundle entries HAPI validates at once. Passed to
+     * {@code FhirValidator.setExecutorService} with concurrent bundle validation enabled,
+     * so only this many InstanceValidator runs are in flight.
      */
     @Getter @Setter
     private int bundleValidationParallelism = 4;
 
-    /**
-     * Entries per chunk. Each chunk is one pool task; HAPI never sees the full bundle.
-     */
-    @Getter @Setter
-    private int bundleValidationBatchSize = 32;
-
     @Bean(name = "bundleValidationExecutor", destroyMethod = "shutdown")
     public ExecutorService bundleValidationExecutor() {
         int parallelism = Math.max(1, bundleValidationParallelism);
-        return Executors.newFixedThreadPool(parallelism, runnable -> {
-            Thread thread = new Thread(runnable, "hapi-bundle-validation");
-            thread.setDaemon(true);
-            return thread;
-        });
+        AtomicInteger threadIndex = new AtomicInteger();
+        return Executors.newFixedThreadPool(parallelism, runnable ->
+                new Thread(runnable, "hapi-bundle-validation-" + threadIndex.incrementAndGet()));
     }
 }
--- a/Java/validation/src/main/java/com/lantanagroup/link/validation/configs/LinkConfig.java
+++ b/Java/validation/src/main/java/com/lantanagroup/link/validation/configs/LinkConfig.java
@@ -11,6 +11,8 @@ import org.springframework.web.client.RestClient;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 @Configuration
 @ConfigurationProperties("link")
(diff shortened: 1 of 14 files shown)


Java/validation/src/main/java/com/lantanagroup/link/validation/configs/LinkConfig.java
```diff
@@ -45,4 +47,27 @@ public ReportClient reportClient(JwtService jwtService, RestClient restClient) {
...
+     */
+    @Getter @Setter
+    private int bundleValidationBatchSize = 32;
+
+    @Bean(name = "bundleValidationExecutor", destroyMethod = "shutdown")
+    public ExecutorService bundleValidationExecutor() {
+        int parallelism = Math.max(1, bundleValidationParallelism);
+        return Executors.newFixedThreadPool(parallelism, runnable -> {
+            Thread thread = new Thread(runnable, "hapi-bundle-validation");
+            thread.setDaemon(true);
```
smailliwcs: Recommend removing this.  We *do* want active bundle-validating threads to prevent the JVM from exiting, so they should be non-daemon.  (Granted, if the executor itself is active, that will also prevent the JVM from exiting, but using daemon threads implies they're performing some kind of low-priority housekeeping task rather than the central aim of the service.)

Java/validation/src/main/java/com/lantanagroup/link/validation/services/ValidationService.java
```diff
@@ -25,37 +26,42 @@
...
 import java.util.Set;
-import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
 
 @Service
-@Scope(value = "prototype", proxyMode = ScopedProxyMode.TARGET_CLASS)
+@Lazy(false)
```
smailliwcs: I think this needs to be reverted.  `ValidationService` must be prototype-scoped to ensure it's using up-to-date artifacts for each `ReadyForValidation` consumption.  An eager-loaded singleton would instead use whatever artifacts were current at startup for the lifetime of the service.  (I.e., updates to artifacts would require a service restart.)

Java/validation/src/main/java/com/lantanagroup/link/validation/services/ValidationService.java
```diff
@@ -25,37 +26,42 @@
...
-        this.validationResultIgnoreService = validationResultIgnoreService;
-
         loadTerminologyValidationSupport(fhirContext, linkConfig, validationSupportChain, validationCacheService);
 
         CachingValidationSupport cachingValidationSupport = new CachingValidationSupport(validationSupportChain);
         IValidatorModule validatorModule = new FhirInstanceValidator(cachingValidationSupport);
         fhirValidator = new FhirValidator(fhirContext);
         fhirValidator.registerValidatorModule(validatorModule);
-        fhirValidator.setConcurrentBundleValidation(true);
-        fhirValidator.setExecutorService(ForkJoinPool.commonPool());
```
smailliwcs: Rather than using custom chunking logic (i.e., the new `validateBundle` path), could we just leave concurrent bundle validation turned on, but use the new `bundleValidationExecutor` bean as the executor service?  In other words, would replacing the common `ForkJoinPool` with `bundleValidationExecutor` resolve the OOM issues?  And if so, do we still need the other changes in this class?

smailliwcs (changes requested):

Java/validation/src/main/java/com/lantanagroup/link/validation/configs/LinkConfig.java
nvmLantana: Removed `setDaemon`. You're right - these threads are doing the actual work of the service, not housekeeping. They're non-daemon now (Thread constructor default) and named `hapi-bundle-validation-N`.

Java/validation/src/main/java/com/lantanagroup/link/validation/services/ValidationService.java
nvmLantana: Reverted. `ValidationService` is prototype-scoped again so each ReadyForValidation picks up current artifacts.

I moved the startup load onto `ArtifactService` instead - it warms `getValidationSupport()` on `ApplicationReadyEvent` so we still don't load IG packages on the first patient while the 18k bundle is already on the heap. save/delete still invalidate that cache, so the next prototype construction sees the update without a restart.

Java/validation/src/main/java/com/lantanagroup/link/validation/services/ValidationService.java
nvmLantana: Yeah, that should be enough. HAPI's concurrent bundle path already validates entries individually and rewrites the `Bundle.entry[n]` locations, so the custom chunking was duplicating that.

I turned concurrent bundle validation back on and pointed it at `bundleValidationExecutor` (4 threads) instead of `ForkJoinPool.commonPool()`. Dropped the chunking path and `bundle-validation-batch-size`. The OOM we hit was mostly lazy package load + a 1g heap with the 18k bundle already in memory - bounding the executor (and sizing the heap from RAM) is the actual concurrency/memory control we needed.

smailliwcs (approved):

PR #1863, opened 2026-08-26
LEGLINK-904: Automation UI: Request/response payload are truncated

SaraMunich-Lantana:
### 🛠️ Description of Changes

Updated API Health diagnostic result handling to prevent request and response payloads from being truncated. Removed the shared payload truncation applied by ServiceTestSuiteBase and the additional response-length limits present in individual API Health test suites so complete request and response bodies are retained for test results and history. Updated the ApiTestRunResult documentation to reflect the full response-body behavior while preserving existing truncation used specifically for exception and error-message diagnostics.

### 🧪 Testing Performed

Successfully built and ran the updated Automation UI. Performed targeted API Health testing to verify that newly generated request and response payloads are retained and displayed in full using the existing scrollable detail view, including comparison with previously persisted results that demonstrated the former truncation behavior. Completed a full Run All API Health regression across the available test suites and confirmed that the request/response truncation issue is resolved.

### 🧑‍🔬 Unit Testing

- Coverage: 7.7%

N/A

### 📓 Documentation Updated

N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * API health test results now preserve complete HTTP response bodies instead of truncating them to 500 characters.
  * Detailed response content is available across service and endpoint checks, improving troubleshooting and diagnosis.
  * Empty or whitespace-only responses continue to display the relevant test note.


nvmLantana (dismissed):

DotNet/Automation.UI/Services/ApiHealth/TestSuites/ServiceTestSuiteBase.cs
```diff
@@ -303,8 +303,7 @@ private static string Truncate(string? value, int maxLength = 300) =>
 
     private static string BodyOrNote(string? value, string note)
     {
-        var body = Truncate(value);
-        return string.IsNullOrWhiteSpace(body) ? note : body;
+        return string.IsNullOrWhiteSpace(value) ? note : value;
```
nvmLantana: This used to `Truncate` at 300. Fine for the live UI, not fine for what `SaveRunResultsAsync` writes.

Those strings land on `ApiHealthRunDocument.EndpointResults` and get upserted as one Cosmos doc per (RunId, ServiceName). Deployed is Cosmos Mongo RU with the **2MB** default (16MB docs are not on). Local Docker Mongo is 16MB, so a laptop run will not catch this.

If `ReplaceOneAsync` throws, `RunSuiteAsync` has already not emitted SSE yet and has no catch, so the rest of the run dies. Keep full bodies in memory/SSE, cap or blob-offload (`InternalBlobStorage`) what we persist. Don't copy the 12MB log-chunk budget.

DotNet/Automation.UI/Services/ApiHealth/TestSuites/DataAcquisitionTestSuite.cs
```diff
@@ -579,9 +579,7 @@ private async Task<ApiTestRunResult> CallRawGetAsync(
 
             result.ResponseBody = string.IsNullOrWhiteSpace(responseBody)
                 ? $"No response body was returned (HTTP {result.ActualStatusCode})."
-                : responseBody.Length > 500
-                    ? responseBody[..500]
-                    : responseBody;
+                : responseBody;
```
nvmLantana: DA is the worst case for this: 60 nested results in one `api_health_runs` doc. Unbounded `responseBody` here (and the same change in Report's 55 steps / MeasureEval GetAll) is how we 413 Cosmos.

Same persist-vs-live split as `BodyOrNote`. Account/Audit/Terminology still do `[..500]` on their GET helpers, so we currently have three policies.

DotNet/Automation.UI/Models/ApiHealth/ApiTestRunResult.cs
```diff
@@ -54,6 +54,6 @@ public sealed class ApiTestRunResult
     /// <summary>Trace ID from the response headers for correlating with Grafana/distributed tracing.</summary>
     public string? TraceId { get; set; }
 
-    /// <summary>Truncated response body (for both success and failure diagnostics).</summary>
+    /// <summary>Response body (for both success and failure diagnostics).</summary>
```
nvmLantana: Comment now reads like this is always the HTTP body. Exception paths in `ServiceTestSuiteBase` still do `Truncate(ex.ToString())` (300 chars). Worth one line in the XML so we don't lie. `ResponseSnippet` is unused; that's probably the persisted cap if we don't want a new field.

nvmLantana (changes requested):
LEGLINK-904's live scrollable view is the right fix, but we now persist untruncated request/response bodies on `ApiHealthRunDocument.EndpointResults` (`MongoApiHealthRunStore.SaveRunResultsAsync`). That's one Cosmos doc per (RunId, ServiceName). DataAcquisition is 60 nested results, Report is 55.

Deployed Automation UI is Cosmos DB for MongoDB RU. 16MB documents are **not** enabled, so the cap is 2MB. Local `mongo:6` is 16MB, so this will not show up on a laptop. Average ~35KB per DA step, or one fat MeasureEval GetAll / Report FHIR payload, and we are on that cliff.

`RunSuiteAsync` saves before it emits SSE `result` events, with no try/catch around the write. A 413 / RU failure kills the rest of the run, skips remaining suites, and we never even get the live UI event for that suite. Index and history also hydrate full bodies into the DOM.

Keep full bodies in the in-memory/SSE result. Cap or blob-offload (`InternalBlobStorage`, same pattern as snapshots) what we persist. Do not copy the 12MB log-chunk budget, and do not use GridFS. Persist-full-only-on-failure is a cheap extra.

Account/Audit/Terminology `CallRawGetAsync` still truncates at 500, so the PR description is a bit incomplete. Pull those into whatever persist-vs-live helper we land on.

DotNet/Automation.UI/Models/ApiHealth/ApiTestRunResult.cs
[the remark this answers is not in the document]
SaraMunich-Lantana: changed it to this:

    /// <summary>Response body (or exception details) for both success and failure diagnostics.</summary>

nvmLantana:
Looked at `c28eada`, `86a53f0`, and `acb4f8aa` against the open threads from me, Copilot, and CodeRabbit. Resolving those threads. This is the current state.

**Fixed**
- Account / Audit / Terminology `CallRawGetAsync` no longer `[..500]`. Capture is consistent across suites.
- `ResponseBody` XML now covers exception details.
- `ApiHealthErrorFormattingTests` asserts 1000-char request/response bodies through `ServiceTestSuiteBase`. That's enough for the capture change. Don't need a copy in every suite.
- `api_health_run_results` (one doc per endpoint result, empty `EndpointResults` on new `api_health_runs` docs) stops DA's 60 nested results (and Report's 55) from stacking in a single Cosmos doc. That's the accumulation problem.

**Not fixed (this is still the original persist-vs-live issue)**
- Each result doc still writes unbounded `RequestBody` / `ResponseBody`. Deployed Automation UI is Cosmos DB for MongoDB RU. Default max document size is **2MB**. `EnableMongo16MBDocumentSupport` is not on. Local `mongo:6` is 16MB, so a laptop run still will not catch this. One fat MeasureEval GetAll or Report FHIR payload still 413s.
- No persist cap, no `InternalBlobStorage` offload, no persist-full-only-on-failure. Keep full bodies in memory/SSE. Cap or blob what we persist. Don't copy the 12MB log-chunk budget, and don't use GridFS. `ResponseSnippet` is still unused if we don't want a new field.
- `RunSuiteAsync` still `await store.SaveRunResultsAsync(...)` before it emits SSE `result` events, with no try/catch. `PersistExecutionStatusAsync` and `CompleteAsync` already swallow write failures. A 413 on one result still kills remaining suites, and we never get the live UI event for that suite.

**New from `acb4f8aa`**
- `GetLatestResultsByServiceAsync` (Index) does `Find(keys.Contains(EndpointKey)).ToListAsync()` then picks latest in memory. That used to be an aggregation for latest-per-service. Dashboard load now pulls every historical full body for every endpoint key.
- `GetHistoryAsync` loads every result for the endpoint, then `Skip`/`Take` in memory. Used to be Mongo skip/limit.
- `MongoIndexManager` never got indexes for `api_health_run_results`. Queries are on `EndpointKey`, `RunId`, and `ServiceName`. Add the same style of indexes we already have on `api_health_runs`.
- No tests for the store rewrite.

Copilot's 16MB Mongo note is the wrong cap (it's Cosmos 2MB). CodeRabbit's GetResults auth/redaction is pre-existing and out of scope for LEGLINK-904. Redacting the bodies would undo the ticket.

nvmLantana (changes requested):
see my previous comment.

nvmLantana (approved):
Looked at 6637620. Indexes are on api_health_run_results at startup (EndpointKey+StartedAt, unique RunId+ServiceName+EndpointKey). Latest-by-service uses the lightweight run docs then fetches only those result docs. History limits the full-body read. Store coverage is in MongoApiHealthRunStoreTests. Approving.

PR #1867, opened 2026-08-27
LEGLINK-1109: Drop empty ResourcesAcquired cache keys and fix DA resource-id SQL timeouts

nvmLantana:
### 🛠️ Description of Changes

LEGLINK-1109. After org-location strip, Data Acquisition still listed empty Encounter cache keys on `ResourcesAcquired`. Normalization retried those as a cache-not-ready miss (LEGLINK-1097), which cannot recreate data that was never cached. Hybrid cache also dropped or split the Redis-vs-ABS decision, so remaining keys missed to Redis while blobs still lived in ABS. `GetResourceIdsForReportPatient` joined unindexed `FhirQuery.FacilityId` and timed out at 30s on Azure SQL. Unproduced tails were not recovered after a failed Kafka produce.

This PR:
- Adds `ResourcesAcquiredTailFinalizer` so the tail drops empty cache keys after strip (checks ABS and Redis), stamps `CacheType` from where remaining keys live (prefer ABS if split), and forgets the in-process Hybrid memo. Worker and `TailMessageRecoveryJob` both run it.
- Normalization dead-letters a listed-but-empty cache key (`DeadLetterException`, producer defect). Zero listed keys still produce `ResourcesNormalized`. Dead-letter releases cache keys; retry paths do not.
- Hybrid persists Redis-vs-ABS as `{correlationId}:__cacheType`, keeps that decision when one type key is deleted, and exposes `HasResourcesAsync` / `ForgetCacheTypeForCorrelationId`.
- Rewrites resource-id lookup from indexed `DataAcquisitionLog` (`FacilityId` + `CorrelationId`), loads resource ids in a follow-up query for the next-eligible batch, and adds `IX_DataAcquisitionLogs_FacilityId_CorrelationId`. Forwards report tracking id into the resource-id query.
- Tail recovery finalizes then republishes; `RevertTailSentAsync` if produce fails.

### 🧪 Testing Performed

- Unit tests: `ResourcesAcquiredTailFinalizerTests`, `ResourcesAcquiredListenerEmptyCacheTests` (dead-letter on listed empty key; produce when no keys listed), `HybridResourceCacheTests` (delete does not forget cache type; persist/replica memo), `ResourceIdParameterFactoryTests` (report tracking id).
- Integration: DataAcquisition fixture / Normalization listener updates for empty-cache tail finalizer and ABS miss retry.

### 🧑‍🔬 Unit Testing

- Coverage: 74.5%

### 📓 Documentation Updated

None. Tail finalizer, Hybrid cache-type persistence, query path, and empty-key dead-letter only.


DotNet/DataAcquisition.Domain/Application/Services/ResourcesAcquiredTailFinalizer.cs
```diff
@@ -0,0 +1,76 @@
...
+            tail.PatientId.SplitReference(),
+            cancellationToken);
+
+        var listed = tail.ResourcesAcquired.CacheKeys ?? [];
+        if (listed.Count == 0)
+        {
+            return;
+        }
+
+        var cache = _resourceCache.GetImplementation(tail.ResourcesAcquired.CacheType) ?? _resourceCache;
```
smailliwcs: The cache type here (populated in `DataAcquisitionLogManager.TryCompleteTailAsync`) ultimately relies on `HybridResourceCache.GetCacheTypeForCorrelationId`, which will always fall back to Redis when called from DA (i.e., from the `TailMessageRecoveryJob`), because DA doesn't write resources to the cache ... only Worker does.  And the implementation of `GetCacheTypeForCorrelationId` relies on the (in-process) `_correlationCacheTypes` to map correlation IDs to cache types.  (And now that I think of it, is that approach broken even within Worker itself when there are multiple replicas?  Each replica will only see the mappings that it has written.)  At any rate, I think the upshot is that anything in ABS will not be `kept` below.

DotNet/Normalization/Listeners/ResourcesAcquiredListener.cs
```diff
@@ -233,13 +233,16 @@ public async Task ProcessMessageAsync(ConsumeResult<ResourceKey, ResourcesAcquir
...
                 {
-                    // Data Acquisition only lists a cache key when it acquired at least one
-                    // resource of that type. An empty read means the cache copy is not ready
-                    // (or used a blob type the reader could not see). Fail transiently so the
-                    // source keys are NOT deleted and MeasureEval does not evaluate an empty bundle.
-                    throw new TransientException(
-                        $"Resource cache key '{cacheKey.SanitizeForLog()}' was listed on ResourcesAcquired but contained no resources. " +
-                        $"CacheType={result.Message.Value.CacheType}, FacilityId={result.Message.Key.FacilityId.SanitizeForLog()}.");
+                    // A listed key with no data is a producer defect (or a still-propagating write).
+                    // Skip this type so a single empty key cannot dead-letter the whole patient.
```
smailliwcs: Don't we *want* the patient to dead-letter if we can't read some resources that DA asserts were acquired?

DotNet/Normalization/Listeners/ResourcesAcquiredListener.cs
```diff
@@ -331,6 +334,14 @@ public async Task ProcessMessageAsync(ConsumeResult<ResourceKey, ResourcesAcquir
                 copiedKeys.Add(cacheKey);
             }
 
+            if (copiedKeys.Count == 0 && cacheKeys.Count > 0)
```
smailliwcs: What if the incoming `ResourcesAcquired` message contains no cache keys?  As it stands, we'll skip this block and go on to produce an "empty" `ResourcesNormalized` message.  Should we dead-letter instead?

DotNet/Shared/Application/Services/ResourceCache/HybridResourceCache.cs
```diff
@@ -96,11 +96,9 @@ public async Task DeleteAsync(List<string> cacheKeys, CancellationToken cancella
                 await cache.DeleteAsync(group.ToList(), cancellationToken);
             }
 
-            // Clean up recorded decisions for deleted keys
-            foreach (var key in cacheKeys)
-            {
-                _correlationCacheTypes.TryRemove(ExtractCorrelationId(key), out _);
-            }
```
smailliwcs: This was the only eviction path for the mappings in `_correlationCacheTypes`.  We'll need some other cleanup mechanism to avoid unbounded growth.

DotNet/DataAcquisition/Jobs/TailMessageRecoveryJob.cs
```diff
@@ -81,6 +82,9 @@ public async Task Execute(IJobExecutionContext context)
                         continue;
                     }
 
+                    var tailFinalizer = scope.ServiceProvider.GetRequiredService<IResourcesAcquiredTailFinalizer>();
+                    await tailFinalizer.FinalizeAsync(tailResult, context.CancellationToken);
```
smailliwcs: Maybe nothing to be done here, but how do we recover if this throws (since we've already marked the tail as having been sent)?  Same issue in `AcquisitionProcessorBackgroundService`.

DotNet/DataAcquisition.Domain/Application/Services/ResourcesAcquiredTailFinalizer.cs
```diff
@@ -0,0 +1,76 @@
...
+        var kept = new List<string>(listed.Count);
+        foreach (var key in listed)
+        {
+            var resources = await cache.GetAsync(key, cancellationToken);
+            if (resources.Count > 0)
+            {
+                kept.Add(key);
+            }
+        }
```
smailliwcs: Does this loop ultimately deserialize all of the patient's resources?  Do we need a faster way to have `IResourceCache` give us a resource count per key without deserializing?

smailliwcs (changes requested):

DotNet/Normalization/Listeners/ResourcesAcquiredListener.cs
nvmLantana: Yeah I thought the same thing, but that assumption breaks down when DA is issuing resource acquired messages for resources that it filters out and therefore doesn't cache it for normalization to read it, but still produces resource acquired. Perhaps I can look at that workflow specifically and see if it can be touched up, and then we can always expect a resource at the cache location in Norm.

DotNet/DataAcquisition.Domain/Application/Services/ResourcesAcquiredTailFinalizer.cs
nvmLantana: Yeah this was a real bug. `GetCacheTypeForCorrelationId` only looked at the in-process dictionary, so DA recovery (and other worker replicas) always defaulted to Redis and would drop ABS keys. I persisted the Redis-vs-ABS decision to Redis at `{correlationId}:__cacheType` (`SET NX`) so any process can resolve it. The tail finalizer also existence-checks both stores (without deserializing) and restamps `CacheType` if the data is actually in ABS. `GetAsync` will try the other store when there's no memo, so strip during recovery doesn't miss ABS encounters.

DotNet/Normalization/Listeners/ResourcesAcquiredListener.cs
nvmLantana: Follow-up after looking at that workflow: Observations/etc. that fail the org-map filter are already dropped *before* we write the cache or record ResourceIds, so they never become cache keys. Encounters are the exception — we have to cache them ungated because location mappings aren't done yet, then strip non-org ones at tail time. ResourceIds still said "Encounter", so we were listing an empty key. The tail finalizer now drops keys that don't actually have resources before we produce, so Norm shouldn't see that case. If a listed key is still empty, that's a producer defect — I changed Norm to dead-letter instead of retry. Retries can't create data we never cached, and they were just bogging the retry topic down.

DotNet/Normalization/Listeners/ResourcesAcquiredListener.cs
nvmLantana: I don't think we should DLQ this one. Zero cache keys means DA finished and there's nothing to normalize (everything stripped, or nothing acquired) — downstream still needs `ResourcesNormalized` so the pipeline can complete. That's "nothing to normalize", not "cache should have been there". I added a log on that path but still produce.

DotNet/Shared/Application/Services/ResourceCache/HybridResourceCache.cs
nvmLantana: Yeah, removing the `TryRemove` on `DeleteAsync` was necessary — org-location strip only deletes Encounter, and Patient/Location get purged by Normalization through `GetImplementation`, so Hybrid never sees a last-key delete. Replaced it with a sliding TTL (same as Redis `CacheEntryTtlDays`) plus `ForgetCacheTypeForCorrelationId` after a successful tail. The Redis `{corr}:__cacheType` memo is left in place so other processes can still resolve the type.

DotNet/DataAcquisition/Jobs/TailMessageRecoveryJob.cs
nvmLantana: Yeah, more of a problem now that `FinalizeAsync` does cache I/O after the claim. Worker and the recovery job now call `RevertTailSentAsync` if finalize/produce throws, and `TryCompleteTailAsync` unclaims if it blows up after claiming. Safety-net can retry.

DotNet/DataAcquisition.Domain/Application/Services/ResourcesAcquiredTailFinalizer.cs
nvmLantana: Yep, it was deserializing everything just to check `Count > 0`. Added `HasResourcesAsync` (Redis `HLEN`, ABS ids-blob exists) and the tail loop uses that instead.

DotNet/Normalization/Listeners/ResourcesAcquiredListener.cs
```diff
@@ -233,11 +233,10 @@ public async Task ProcessMessageAsync(ConsumeResult<ResourceKey, ResourcesAcquir
...
-                    // resource of that type. An empty read means the cache copy is not ready
-                    // (or used a blob type the reader could not see). Fail transiently so the
-                    // source keys are NOT deleted and MeasureEval does not evaluate an empty bundle.
-                    throw new TransientException(
+                    // DA only lists a key after it has written (and not stripped) resources there.
+                    // An empty listed key is a producer defect, not a not-ready race: retries cannot
+                    // create data that was never cached (org-map filter, Encounter strip, etc.).
+                    throw new DeadLetterException(
                         $"Resource cache key '{cacheKey.SanitizeForLog()}' was listed on ResourcesAcquired but contained no resources. " +
                         $"CacheType={result.Message.Value.CacheType}, FacilityId={result.Message.Key.FacilityId.SanitizeForLog()}.");
```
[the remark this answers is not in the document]
nvmLantana: PR description is out of date. Will update the descripton.

DotNet/DataAcquisition.Domain/Application/Services/ResourcesAcquiredTailFinalizer.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        // Prefer ABS when a correlation has data in both stores (a replica split) so
+        // Normalization does not look only in Redis and miss the ABS payload.
+        if (sawAbs)
+        {
+            tail.ResourcesAcquired.CacheType = ResourceCacheType.ABS;
+        }
+        else if (sawRedis)
+        {
+            tail.ResourcesAcquired.CacheType = ResourceCacheType.Redis;
```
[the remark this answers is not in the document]
nvmLantana: Yeah, mixed Redis/ABS keys would stamp ABS and leave Redis-only payloads unread. Finalizer now copies Redis-only keys into ABS, deletes the Redis copies of kept keys, then stamps CacheType ABS. Covered by FinalizeAsync_CopiesRedisOnlyKeysIntoAbsWhenStoresAreSplit.

DotNet/DataAcquisition.Domain/Application/Managers/DataAcquisitionLogManager.cs
```diff
@@ -880,69 +886,105 @@ public async Task<int> StampSiblingCountAsync(string facilityId, string correlat
...
+                        .Select(rt => $"{groupInfo.CorrelationId}:{rt}")
+                        .Distinct()
+                        .ToList()
+                }
+            };
+        }
+        catch
+        {
+            await RevertTailSentAsync(groupInfo.FacilityId, groupInfo.CorrelationId!, groupInfo.QueryPhase, CancellationToken.None);
+            throw;
```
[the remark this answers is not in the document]
nvmLantana: Yeah, if RevertTailSent failed after TailSent was already true, GetOrphanedTailLogIds could never pick the group back up. TryCompleteTailAsync now claims TailClaimedAt and only sets TailSent after Kafka produce. Stale claims get reclaimed after the 15 minute lease. Covered by TryCompleteTailAsync_WhenClaimIsStale_Reclaims and Execute_ReclaimsStaleTailClaim_ProducesAndMarksSent.

smailliwcs (approved):

PR #1868, opened 2026-08-27
LEGLINK-1110: Score ORM reuse against raw Locations, not cleanup type codes

nvmLantana:
### 🛠️ Description of Changes

LEGLINK-1110. The ORM suggestor scored reuse as if Normalization cleanup type codes (Copy Location, alias copy, Location CodeMap) would already be on the Location. Org mapping runs during acquisition on raw Locations, so type-only maps were recommended Reuse when they would not match.

This PR scores ORM reuse against identifiers and type codes already on the uploaded Locations. Type-only maps are Reuse only when those types are already present. Value-specific identifier conditions no longer cover the whole system. Reuse ranks above Extend. Normalization proposals that rewrite Location.type note that org maps must match the raw resource.

### 🧪 Testing Performed

- Unit tests in `BundleConfigurationGenerationTests`: type-only map not reused when the upload has identifiers but not those types; type-only map reused when those types are already on the upload; value-specific identifier condition does not cover the whole system; Normalization note that org maps must match raw Locations.

### 🧑‍🔬 Unit Testing

- Coverage: 90.8%

### 📓 Documentation Updated

None. Suggestor scoring and notes only.

### DIFF
--- a/DotNet/Automation.UI/Services/ConfigurationGeneration/OrgResourceMapProposalBuilder.cs
+++ b/DotNet/Automation.UI/Services/ConfigurationGeneration/OrgResourceMapProposalBuilder.cs
@@ -23,7 +23,7 @@ public static class OrgResourceMapProposalBuilder
             SuggestedName = refineExisting is { IsSystem: false }
                 ? refineExisting.Name
                 : "",
-            SuggestedDescription = "Auto-built so every Location found in the uploaded bundle can pass org-location mapping (any-match)."
+            SuggestedDescription = "Auto-built so every Location can pass org mapping during acquisition (any-match), using identifiers or type codes already on the raw resource — not codes cleanup adds later."
         };
 
         var conditions = new List<OrganizationResourceMapCondition>();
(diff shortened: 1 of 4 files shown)


DotNet/Automation.UI/Services/ConfigurationGeneration/OrgResourceMapProposalBuilder.cs
```diff
@@ -99,48 +106,119 @@ public static List<ReuseCandidate> ScoreExisting(
...
-        var systems = DistinctIdentifierSystems(fingerprint);
-        foreach (var system in systems)
-            keys.Add($"idsys|{system}");
+        foreach (var identifier in fingerprint.LocationIdentifiers)
+        {
+            var system = identifier.System?.Trim() ?? "";
+            var value = identifier.Value?.Trim() ?? "";
+            if (string.IsNullOrWhiteSpace(system))
+                continue;
+            keys.Add(string.IsNullOrWhiteSpace(value)
```
SaraMunich-Lantana: Could we add a unit test confirming that a system-wide identifier condition still receives Reuse when the uploaded Location has a specific value for that system? Since NeededIdentifierKeys() can now include both idsys|system and id|system|value, I want to make sure a broader idsys condition is considered to cover the value-specific key rather than being scored as partial coverage and downgraded to Extend.

nvmLantana: This potentiality is already implicitly covered through the existing test Orm_builder_recognizes_system_default_or_style_path

SaraMunich-Lantana (approved):
Looks good

PR #1871, opened 2026-08-28
LEGLINK-204: Update Automation With Thetis and Create Metrics Dashboard

nvmLantana:
### 🛠️ Description of Changes

Please provide a high-level overview of the changes included in this PR.

### 🧪 Testing Performed

### 🧑‍🔬 Unit Testing

- Coverage: 55.7%

### 📓 Documentation Updated


DotNet/Automation.UI/Dockerfile
```diff
@@ -5,20 +5,31 @@ FROM lantanagroup/aspnet8-run AS base
...
 COPY ["DotNet/Automation.UI/Automation.UI.csproj", "DotNet/Automation.UI/"]
+ARG FEED_ACCESSTOKEN
 RUN --mount=type=cache,id=nuget,sharing=locked,target=/root/.nuget/packages \
+    if [ -n "$FEED_ACCESSTOKEN" ]; then \
+      rm -rf /root/.nuget/packages/lantanagroup.thetis.generation.abstractions \
+             /root/.nuget/packages/lantanagroup.thetis.generation.engine; \
+      dotnet nuget update source Shared_BOTW_Feed \
+        --username docker \
+        --password "$FEED_ACCESSTOKEN" \
+        --store-password-in-clear-text \
```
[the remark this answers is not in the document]
nvmLantana: CI bake and compose now mount AZURE_ARTIFACTS_PAT as a BuildKit secret (feed_accesstoken). Restore uses a throwaway nuget.config.feed that gets deleted in the same layer. ARG FEED_ACCESSTOKEN is only a local fallback when the secret isn't mounted. Didn't add another test for it.

DotNet/MockFhirServer/Dockerfile
```diff
@@ -4,15 +4,26 @@ RUN apt-get update && apt-get install -y curl && rm -rf /var/lib/apt/lists/*
...
 COPY ["DotNet/MockFhirServer/MockFhirServer.csproj", "DotNet/MockFhirServer/"]
+ARG FEED_ACCESSTOKEN
 RUN --mount=type=cache,id=nuget,sharing=locked,target=/root/.nuget/packages \
+    if [ -n "$FEED_ACCESSTOKEN" ]; then \
+      rm -rf /root/.nuget/packages/lantanagroup.thetis.generation.abstractions \
+             /root/.nuget/packages/lantanagroup.thetis.generation.engine; \
+      dotnet nuget update source Shared_BOTW_Feed \
+        --username docker \
+        --password "$FEED_ACCESSTOKEN" \
+        --store-password-in-clear-text \
```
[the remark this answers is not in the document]
nvmLantana: CI bake and compose now mount AZURE_ARTIFACTS_PAT as a BuildKit secret (feed_accesstoken). Restore uses a throwaway nuget.config.feed that gets deleted in the same layer. ARG FEED_ACCESSTOKEN is only a local fallback when the secret isn't mounted. Didn't add another test for it.

DotNet/Automation.UI/Services/MetricsScenarioFingerprint.cs
```diff
@@ -0,0 +1,59 @@
...
+        int patientCount,
+        int seed,
+        int resourcesMin,
+        int resourcesMax,
+        int? concurrency,
+        string? benchmarkKey,
+        IEnumerable<string>? measures,
+        string? thetisGitSha,
+        Guid? queryPlanId,
+        Guid? normalizationSuiteId)
```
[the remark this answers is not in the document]
nvmLantana: Fingerprint now includes selected measure template ids plus a SHA of the effective bundle JSON. Same generation family with different templates or CQL gets a different version. Covered by Different_measure_template_ids_produce_a_new_fingerprint and Different_measure_bundle_hashes_produce_a_new_fingerprint.

DotNet/Automation/Generation/GeneratedPatientBundleJson.cs
```diff
@@ -0,0 +1,50 @@
...
+        using (var writer = new Utf8JsonWriter(stream, new JsonWriterOptions { Indented = true }))
+        {
+            writer.WriteStartObject();
+            writer.WriteString("resourceType", "Bundle");
+            writer.WriteString("type", "collection");
+            writer.WriteNumber("total", entries.Count);
+            writer.WritePropertyName("entry");
+            writer.WriteStartArray();
+            foreach (var entry in entries)
+                entry.WriteTo(writer);
```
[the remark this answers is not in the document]
nvmLantana: Collection merge now drops entry.request and Bundle.total. MergeToCollection_strips_transaction_request_from_entries covers it.

DotNet/DataAcquisition.AcquisitionWorker/Listeners/ReadyToAcquireListener.cs
```diff
@@ -88,7 +89,8 @@ await logManager.TrySetLogStatusAsync(
         {
             await processor.EnqueueAsync(new AcquisitionWorkItem(
                 LogId: logId,
-                FacilityId: value.FacilityId
+                FacilityId: value.FacilityId,
+                IsPerformanceMode: KafkaHeaderHelper.IsPerformanceMode(consumeResult.Message?.Headers)
```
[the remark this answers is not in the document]
nvmLantana: Metrics dashboard runs are adhoc. DataAcquisitionRequested already carries X-Metrics-Mode, but ReadyToAcquire is produced later from the log table, which has no column for it. Stashing that on DataAcquisitionLog is a schema change. Not doing that in this PR.

DotNet/Report/Business/Managers/ReportEntryManager.cs
```diff
@@ -397,6 +403,10 @@ public async Task<ReportEntryModel> UpdateAsync(ReportEntryModel model, Cancella
             entity.ModifyDate = DateTime.UtcNow;
             entity.AggregateReportUri = model.AggregateReportUri;
             entity.AggregateReportBlobName = model.AggregateReportBlobName;
+            if (entity.ReportingStatus != model.ReportingStatus)
+            {
+                _metrics.IncrementStatusTransition(entity.FacilityId, entity.ReportingStatus.ToString(), model.ReportingStatus.ToString());
+            }
```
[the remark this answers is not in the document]
nvmLantana: IncrementStatusTransition now runs after PersistAsync succeeds, for both reporting-status and measure-status. Didn't add another test for it.

DotNet/Report/Listeners/ReportScheduledListener.cs
```diff
@@ -130,6 +130,7 @@ public async Task ProcessMessageAsync(ConsumeResult<string, ReportScheduledValue
                     throw new DeadLetterException("ReportScheduled event is null.");
                 }
 
+                using var metricsMode = MetricsModeScope.Begin(KafkaHeaderHelper.IsPerformanceMode(result.Message?.Headers));
```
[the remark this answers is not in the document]
nvmLantana: Metrics dashboard runs go through adhoc GenerateReportListener, which already copies inboundMetricsMode onto later produces. ReportScheduled is the census scheduler. Persisting the header on ReportSchedule is a migration. Not in this PR.

DotNet/Automation.UI/Services/TerminologyCodeLookup.cs
```diff
@@ -0,0 +1,122 @@
...
+                using var response = await client.GetAsync(url, cancellationToken);
+                if (!response.IsSuccessStatusCode)
+                    continue;
+                var json = await response.Content.ReadAsStringAsync(cancellationToken);
+                var display = ReadDisplay(json) ?? code;
+                return Build(kind, code, sys, display, incomplete: kind == GenerationCatalogKind.Observation);
+            }
+            catch (Exception ex)
+            {
+                logger.LogDebug(ex, "Terminology lookup failed for {System}|{Code}.", sys, code);
```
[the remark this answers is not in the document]
nvmLantana: Debug log no longer interpolates system or code. The lookup URL still uses Uri.EscapeDataString.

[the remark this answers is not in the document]
nvmLantana: Debug log no longer interpolates system or code. The lookup URL still uses Uri.EscapeDataString.

DotNet/Automation/Generation/Thetis/ThetisPatientEntryGenerator.cs
```diff
@@ -0,0 +1,282 @@
...
+                }
+            },
+            new Encounter.LocationComponent
+            {
+                Location = new ResourceReference($"Location/{ids.IcuLocation}") { Display = "Intensive Care Unit" },
+                Status = Encounter.EncounterLocationStatus.Completed,
+                Period = new Period
+                {
+                    StartElement = new FhirDateTime(encStart.AddHours(4)),
+                    EndElement = new FhirDateTime(encEnd.AddDays(-1))
```
nvmLantana: **[bug]** `AlignEncounterToSharedStay` always overwrites the Thetis Encounter with a three-ward ED to ICU to step-down stay. ICU end and step-down start are `encEnd.AddDays(-1)`. For any stay shorter than about 28 hours (outpatient 2-6h windows, `DurationMinutes`, daily scheduled patterns such as AdmittedDuringPeriodDischargedDuringPeriod) that produces `Period.start > Period.end`. Ambulatory intents still get completed ICU/step-down locations (`Ambulatory_intent_does_not_emit_inpatient_encounter_type` never asserts locations). Location-mapping and encounter-period CQL then see invalid or non-AMB geography.

**Suggestion:** Scale the three location periods to `[encStart, encEnd]` (and skip ICU/step-down for AMB/EMER), or only apply this overlay when `RequiresInpatientEncounter()` and `encEnd - encStart` is long enough for three non-empty intervals.

DotNet/DataAcquisition.AcquisitionWorker/Listeners/ReadyToAcquireListener.cs
nvmLantana: **[bug]** The worker now gates semaphore-wait (and `MetricsModeScope` for the rest of the work item) on `X-Metrics-Mode` from ReadyToAcquire. `AcquisitionProcessingJob` still builds ReadyToAcquire headers with only `X-Correlation-Id`, and `DataAcquisitionLogService` produces ReadyToAcquire with no headers at all. Adhoc metrics runs do put `X-Metrics-Mode=performance` on GenerateReport / DataAcquisitionRequested, but that marker is lost at the log-table hop, so `IsPerformanceMode` is false for every normal metrics run. `AcquisitionProcessorBackgroundService` also produces ResourcesAcquired without copying the mode, so Normalization / MeasureEval / Validation never see it either. QueryDispatch's DataAcquisitionRequested produce is likewise header-only correlation. Copilot raised this; it is still true on this HEAD. Dashboard stage histograms that are always-on can still populate; the new performance-only meters (semaphore wait, terminology lookup duration, report persist after the Report-listener scope) will not.

**Suggestion:** Carry the mode without a SQL migration: stash it on the ReadyToAcquire message (body or header) when DataAcquisitionRequested is consumed, or keep an in-memory/Redis map keyed by facility+reportTrackingId. On the worker, `KafkaHeaderHelper.ApplyIfPerformance` onto ResourcesAcquired from `item.IsPerformanceMode` so downstream copy helpers have something to copy. Add a test that a performance DataAcquisitionRequested results in ReadyToAcquire + ResourcesAcquired carrying `X-Metrics-Mode`.

DotNet/Automation.UI/Services/RunMetricsSnapshotService.cs
```diff
@@ -0,0 +1,665 @@
...
+
+    private async Task<StageLatencySnapshot> QueryStageAsync(
+        StageQuery stage,
+        string facilityId,
+        DateTimeOffset evaluationTime,
+        CancellationToken cancellationToken)
+    {
+        var facility = EscapePromLabel(facilityId);
+        var bucketSelector = $"{stage.HistogramBase}_bucket{{facility_id=\"{facility}\"}}";
+        var countSelector = $"sum({stage.HistogramBase}_count{{facility_id=\"{facility}\"}})";
```
nvmLantana: **[bug]** Stage p50/p95/p99 queries use cumulative `*_count` / `*_bucket` filtered only by `facility_id`, not `increase(...[window])` over the pipeline window (HTTP/CPU queries in the same class do use `increase`/`rate`). Non-DMRP runs use a unique run GUID as facility id, so this happens to be per-run. DMRP metrics runs use the reusable NHSN Organization ID (`state.Options.NhsnOrganizationId`), so a second metrics run against the same org folds every prior histogram observation into the snapshot and can fail a time budget that the current run actually met. Process RSS/CPU in the same capture is also process-wide; overlapping metrics runs contaminate utilization even with unique facility ids.

**Suggestion:** Query `increase(link_*_duration_milliseconds_bucket{facility_id="..."}[<lookback>s])` (and the matching `_count`) at the evaluation timestamp, same lookback used for utilization. For DMRP, prefer a run-scoped tag or record the capture window so successive org-id runs are not compared as one series.

DotNet/Automation.UI/Services/ThetisRevision.cs
```diff
@@ -0,0 +1,76 @@
+using System.Diagnostics;
+using System.Reflection;
+
+namespace Automation.UI.Services;
+
+public static class ThetisRevision
+{
+    public static string? TryGetGitSha()
```
nvmLantana: **[bug]** Scenario fingerprints hash `ThetisRevision.TryGetGitSha()`, which shells out to `git rev-parse` in a sibling `thetis` checkout. Docker/CI images have neither git nor that checkout, so the SHA is always null there. `TryGetAssemblyInformationalVersion()` is stored on the snapshot but is not part of `MetricsScenarioFingerprint.Compute`. Two container runs that differ only by Thetis package revision get the same fingerprint and are treated as comparable / same setup version. `Source = "sibling-project-ref"` on the snapshot is also wrong for NuGet-fed builds.

**Suggestion:** Fingerprint the loaded `Thetis.Generation.Engine` informational version (and/or the lock-file version). Keep git SHA as an optional extra when a sibling repo exists. Stop labeling NuGet builds as `sibling-project-ref`.

docker-compose.yml
```diff
@@ -729,6 +734,9 @@ services:
       dockerfile: DotNet/Automation.UI/Dockerfile
       args:
```
nvmLantana: **[bug]** Compose still passes `FEED_ACCESSTOKEN: ${AZURE_ARTIFACTS_PAT:-}` as a Docker build ARG for automation-ui and mock-fhir-server, and both Dockerfiles fall back to `$FEED_ACCESSTOKEN` when the BuildKit secret is missing (`if [ -z "$token" ]; then token="$FEED_ACCESSTOKEN"; fi`). Build-args are stored in image history/metadata even when the RUN also mounts a secret. Copilot asked for secret-only transport; the secret mount is present, but the ARG path is still live for the compose/CI bake that sets it.

**Suggestion:** Drop `FEED_ACCESSTOKEN` from compose `args` and from the Dockerfiles. Restore only from `--mount=type=secret,id=feed_accesstoken`. For local non-CI builds, document `docker build --secret id=feed_accesstoken,env=AZURE_ARTIFACTS_PAT` (or a user nuget.config) instead of ARG.

nuget.config
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <packageSources>
+    <clear />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+    <add key="Shared_BOTW_Feed" value="https://pkgs.dev.azure.com/lantanagroup/nhsnlink/_packaging/Shared_BOTW_Feed/nuget/v3/index.json" />
```
nvmLantana: **[suggestion]** Restore of Automation / Automation.UI / MockFhirServer now requires `LantanaGroup.Thetis.*` from Azure Artifacts `Shared_BOTW_Feed`. There are no credentials in nuget.config. A developer who currently `dotnet restore`s without `AZURE_ARTIFACTS_PAT` (and without `dotnet nuget update source`) will get NU1301/NU1101 on Thetis. That is a new mandatory local-dev/CI setup step, not just an extra package.

**Suggestion:** Document the PAT (or Azure Artifacts login) in Automation README, and fail the restore with a clear message when the feed 401s. Consider a public/symbol-package fallback for contributors who only build non-Automation services.

DotNet/Automation.UI/Services/MetricsScenarioFingerprint.cs
```diff
@@ -0,0 +1,72 @@
...
+        int? concurrency,
+        string? benchmarkKey,
+        IEnumerable<string>? measures,
+        string? thetisGitSha,
+        Guid? queryPlanId,
+        Guid? normalizationSuiteId,
+        IEnumerable<string>? measureTemplateIds = null,
+        string? measureContentHash = null)
+    {
+        var payload = string.Join('|',
```
nvmLantana: **[suggestion]** After the template-id/bundle-hash fix, the fingerprint still ignores patient configuration identity (cohort template ids, intent palettes, stay pattern, imported mix). Two metrics runs with the same patient count/seed/resources/measures but different clinical shapes share a fingerprint and are compared as regressions.

**Suggestion:** Include ordered patient-configuration ids (or a hash of each cohort's intent JSON + `ScheduledInpatientPattern`) in `Compute`, and cover it with a test next to `Different_measure_bundle_hashes_produce_a_new_fingerprint`.

DotNet/Automation.UI/Controllers/MeasuresController.cs
```diff
@@ -0,0 +1,143 @@
...
+            MeasureBundleParser.ApplyMetadata(model, parsed);
+        }
+        catch (InvalidOperationException ex)
+        {
+            return BadRequest(ex.Message);
+        }
+
+        model.IsSystem = false;
+        model.UpdatedAt = DateTimeOffset.UtcNow;
+        await store.UpsertAsync(model, ct);
```
nvmLantana: **[suggestion]** `SaveInline` upserts `BundleJson` with no size cap. `MongoMeasureTemplateStore` stores that JSON inline and comments that ACH files are "well under the 2 MB Cosmos document limit." Custom uploads (or a future fatter NHSN bundle) will fail the Cosmos 2 MB practical / 16 MB hard document limit with a driver error rather than a 400.

**Suggestion:** Reject (or spill to blob, like generated templates) when `Encoding.UTF8.GetByteCount(model.BundleJson)` exceeds a documented budget (e.g. 1.5 MB) before `UpsertAsync`.

DotNet/Automation/Generation/FhirGenerationPipeline.cs
```diff
@@ -535,11 +572,23 @@ private static (DateTime? Start, DateTime? End) ParseClinicalPeriod(AcquisitionS
 
         if (cachedTemplate == null)
         {
-            // Generate entries using the same per-patient logic shared with FhirBundleGenerator.
-            entries = GeneratePatientEntries(
-                profile, patientIndex, baseSeed, effectiveResourcesPerPatient,
-                sharedPractitionerIds, sharedMedicationIds, measures,
-                generationClinicalPeriodStart, generationClinicalPeriodEnd, config, generationRequirementsPlan, ids);
+            var generator = patientEntryGenerator ?? ThetisPatientEntryGenerator.Shared;
+            entries = await generator.GenerateAsync(new PatientEntryRequest
```
nvmLantana: **[suggestion]** The new Thetis `GenerateAsync(..., CancellationToken)` is invoked without a token. `GenerateAndUploadAsync` still has no CT parameter, and the per-patient `Task.Run` does not take `state.RunCancellation`. UI cancel during a mega-patient Thetis run cannot stop in-flight engine work; cancel is only observed after generation returns.

**Suggestion:** Thread `CancellationToken` through `GenerateAndUploadAsync` / `GenerateAndUploadSinglePatientAsync` into `generator.GenerateAsync` and `semaphore.WaitAsync`, and pass `state.RunCancellation.Token` from `RunExecutor`.

nvmLantana (commented):
## Summary

Thetis-backed generation, reusable measure/patient templates, and a Prometheus-backed metrics dashboard are a coherent slice of work, and several Copilot items (collection-bundle stripping, fingerprint template hashes, ReportEntryManager transition counters, BuildKit secret mounts) look addressed on this HEAD. The remaining defects are concentrated in three places: generated Encounter location periods that invert on short/AMB/daily stays, X-Metrics-Mode dropping at ReadyToAcquire so the new performance-only instruments never fire on the acquisition path, and metrics snapshots that mix prior-run histogram totals plus a Thetis identity that is empty in containers.

## Issue counts by severity

- bugs: 5
- suggestions: 4
- nits: 0

DotNet/Automation.UI/Services/RunMetricsSnapshotService.cs
nvmLantana: System.Collections.Hashtable[4041312852]

DotNet/Automation.UI/Services/MetricsScenarioFingerprint.cs
nvmLantana: System.Collections.Hashtable[4041312876]

DotNet/Automation.UI/Controllers/MeasuresController.cs
nvmLantana: System.Collections.Hashtable[4041312884]

DotNet/Automation/Generation/FhirGenerationPipeline.cs
nvmLantana: System.Collections.Hashtable[4041312889]

DotNet/DataAcquisition.AcquisitionWorker/Listeners/ReadyToAcquireListener.cs
nvmLantana: System.Collections.Hashtable[4041312848]

docker-compose.yml
nvmLantana: System.Collections.Hashtable[4041312864]

DotNet/Automation.UI/Services/ThetisRevision.cs
nvmLantana: System.Collections.Hashtable[4041312859]

DotNet/Automation/Generation/Thetis/ThetisPatientEntryGenerator.cs
nvmLantana: System.Collections.Hashtable[4041312840]

nuget.config
nvmLantana: System.Collections.Hashtable[4041312872]

DotNet/Automation/Generation/Thetis/ThetisPatientEntryGenerator.cs
nvmLantana: Scaled the IP overlay to [encStart, encEnd] so ICU/step-down periods cannot invert on short stays. AMB/EMER gets a single ED location, no ICU/step-down. Covered by Short_inpatient_stay_keeps_forward_location_periods and Ambulatory_intent_does_not_emit_inpatient_encounter_type.

DotNet/DataAcquisition.AcquisitionWorker/Listeners/ReadyToAcquireListener.cs
nvmLantana: No SQL column. DataAcquisitionRequestedListener stashes performance mode in ICacheService keyed by facility+reportTrackingId. AcquisitionProcessingJob copies it onto ReadyToAcquire headers. The worker then ApplyIfPerformance onto ResourcesAcquired. ReportMetricsModeCacheTests and ExecuteListenerAsync_WhenPerformanceHeader_EnqueuesWorkItemInPerformanceMode cover the hops.

DotNet/Automation.UI/Services/RunMetricsSnapshotService.cs
nvmLantana: Stage p50/p95/p99 and counts now use increase() over the same lookback as utilization. Stage_queries_use_increase_over_the_window covers the PromQL. Process RSS/CPU is still process-wide. Did not add a run-scoped tag for that.

DotNet/Automation.UI/Services/ThetisRevision.cs
nvmLantana: Fingerprint now includes Thetis.Generation.Engine informational version. Git SHA stays optional when a sibling checkout exists. Snapshot Source is nuget vs sibling-project-ref vs unknown. Different_thetis_versions_produce_a_new_fingerprint covers it.

docker-compose.yml
nvmLantana: Cannot drop ARG on this runner. Self-hosted docker buildx bake rejects --secret (unknown flag) and --set *.secrets. Compose still mounts feed_accesstoken for compose-build. ARG FEED_ACCESSTOKEN is the bake path. Did not add another test for it.

nuget.config
nvmLantana: Documented the Shared_BOTW_Feed PAT restore in Automation and Automation.UI READMEs. packageSourceMapping already keeps LantanaGroup.Thetis.* on that feed and everything else on nuget.org. No public Thetis fallback.

DotNet/Automation.UI/Services/MetricsScenarioFingerprint.cs
nvmLantana: Fingerprint now includes ordered patient-configuration / intent shape keys (config id, stay pattern, class, duration, insulin, primary dx, count). Different_patient_shape_keys_produce_a_new_fingerprint covers it.

DotNet/Automation.UI/Controllers/MeasuresController.cs
nvmLantana: SaveInline rejects BundleJson over 1.5 MB before UpsertAsync. SaveInline_rejects_bundle_over_size_cap covers it.

DotNet/Automation/Generation/FhirGenerationPipeline.cs
nvmLantana: GenerateAndUploadAsync takes CancellationToken now. It goes into semaphore.WaitAsync and generator.GenerateAsync. RunExecutor passes state.RunCancellation.Token. Did not add another test for it.

DotNet/Automation/README.md
```diff
@@ -5,6 +5,14 @@ generation, predictive pipeline modeling, and reusable automation primitives. It
 dependency on any Link-specific service; host projects compose it with environment-specific
 orchestration.
 
+Thetis generation packages (`LantanaGroup.Thetis.Generation.Abstractions` / `Engine`) come from Azure Artifacts feed `Shared_BOTW_Feed`. Authenticate once per machine before `dotnet restore` of Automation / Automation.UI / MockFhirServer:
```
johnbritton: I suggest adding these instructions to DEVELOPMENT.md. Also add instructions to generate an Azure DevOps personal access token scoped to READ access for packaging. I also needed to add an environment variable called `AZURE_ARTIFACTS_PAT` for the local docker-compose.

Java/validation/src/main/java/com/lantanagroup/link/validation/services/ReadyForValidationConsumer.java
```diff
@@ -203,9 +212,10 @@ private List<Result> validate(String correlationId, String facilityId, String pa
             results = validationService.validate(bundle, facilityId, reportId);
             _logger.debug("Validation completed with {} results in {} seconds", results.size(), String.format("%.2f", timer.getSeconds()));
 
-            attributes = buildMetricAttributes(bundle, results, correlationId, facilityId, patientId, reportId);
+            attributes = buildMetricAttributes(results, facilityId);
```
johnbritton: Possibly an issue with metric accuracy:

Validation metrics are calculated before categorization. At `ReadyForValidationConsumer.java:215`, issue counts see categorized results as uncategorized because categorization only occurs at `ReadyForValidationConsumer.java:230`. This makes validation outcomes and severity metrics incorrect.

Java/validation/src/main/java/com/lantanagroup/link/validation/services/ReadyForValidationConsumer.java
```diff
@@ -279,6 +289,16 @@ private void produceValidationCompleteRecord(
             String patientId,
             String reportId,
             List<Result> results) {
+        produceValidationCompleteRecord(correlationId, facilityId, patientId, reportId, results, null);
```
johnbritton: Not sure if uncategorized is the same as a null category. Leaving it here just in case.

Uncategorized results can crash completion processing. `countIssuesBySeverity()` explicitly supports null categories, but `produceValidationCompleteRecord()` unconditionally calls `result.getCategories().stream()` at `ReadyForValidationConsumer.java:305-306`. A validation result with no category can therefore fail the Kafka consumer instead of producing `ValidationComplete`.

DotNet/Automation/Generation/FhirGenerationPipeline.cs
```diff
@@ -754,110 +774,187 @@ private static (DateTime? Start, DateTime? End) ParseClinicalPeriod(AcquisitionS
...
-                    ids.EdLocation, ids.IcuLocation,
-                    ids.StepDownLocation, ids.Organization,
-                    anchors.PrimaryDxId, scenario);
+                var bundles = ResolveMeasureBundles(measureBundleJsons, measures, qualifyingMeasures);
+                cqlFilteredKeys = CqlFilterSimulator.ComputeFilteredKeys(bundles, cqlInput);
+                manifestBuilder.SetCqlFilteredKeys(patientId, cqlFilteredKeys);
             }
         }
-        else
+
```
johnbritton: No idea if this one is legit, but it sounds like a future headache if true.

Parallel patient workers append `PatientProfile` entries in completion order, while `PatientIds` and later manifest logic assume profile order. With concurrent generation, profiles can be paired with the wrong patient, corrupting qualification and expected ABS counts. Store results by `patientIndex` and build the manifest deterministically afterward.

DotNet/Automation/Generation/FhirGenerationPipeline.cs
```diff
@@ -754,110 +774,187 @@ private static (DateTime? Start, DateTime? End) ParseClinicalPeriod(AcquisitionS
...
+            var acquiredKeys = QueryPlanAcquisitionSimulator.SimulateAcquiredKeysForPatient(
+                patientId,
+                patientSimEntries,
+                sharedSimEntries,
+                acquisitionSimulation.QueryPlan,
+                acquisitionSimulation.ClinicalPeriodStart,
+                acquisitionSimulation.ClinicalPeriodEnd,
+                output,
+                acquisitionSimulation.AllowEncounterAnchoredDateOverrideForOutOfRange);
+            acquiredKeys = OrgResourceMapPredictionFilter.Apply(
```
johnbritton: This could break tests with org mapping, if true.

The refactor applies organization-location filtering after CQL filtering and no longer restricts CQL encounter inputs to organization-acquired encounters. A patient whose only qualifying encounter is removed by the org map can remain marked qualifying, producing false ABS/report expectations. Restore the previous acquired-encounter restriction before calculating eligibility and CQL-filtered keys.

johnbritton (changes requested):
I'll let you determine if the code issues are legit, but I would like to see better docs around the new requirement for a PAT.

DotNet/Automation/README.md
nvmLantana: PAT create steps and AZURE_ARTIFACTS_PAT for compose are in DEVELOPMENT.md now. Automation README points there. Visual Studio can still sign into the lantanagroup Azure DevOps org instead of using a PAT, if the account has Read on Shared_BOTW_Feed.

Java/validation/src/main/java/com/lantanagroup/link/validation/services/ReadyForValidationConsumer.java
nvmLantana: Issue counts and validation.outcome were taken before categorize(), so every HAPI result looked uncategorized. addToValidationCounter and addIssues now run after categorizationService.categorize. Validation duration still times validate() only. process_recordsIssueMetricsAfterCategorize covers it.

Java/validation/src/main/java/com/lantanagroup/link/validation/services/ReadyForValidationConsumer.java
nvmLantana: produceValidationCompleteRecord now treats a null category list as empty instead of calling stream() on it. Categorize already writes an empty list in production. process_nullCategories_stillProducesValidationComplete covers the null path.

DotNet/Automation/Generation/FhirGenerationPipeline.cs
nvmLantana: Did not change this. IncrementalBuilder.AddPatient locks and adds the patient id and profile together, so they stay paired. Concurrent workers write patientIds[patientIndex], not append in completion order. Manifest lookups are by patient id.

DotNet/Automation/Generation/FhirGenerationPipeline.cs
nvmLantana: Did not change this. OrgResourceMapPredictionFilter prunes simulated acquired keys after CQL. Q/NQ still uses every generated encounter in the measurement period. Restricting CQL input to org-mapped encounters would change eligibility for org-mapped facilities. Not in this PR.

DotNet/Automation/README.md
nvmLantana: DEVELOPMENT.md now also notes you can add Shared_BOTW_Feed as a NuGet source in Visual Studio (Tools > Options > NuGet Package Manager > Package Sources) and sign into the lantanagroup Azure DevOps org instead of using a PAT.

johnbritton (approved):

DotNet/Automation/Generation/FhirGenerationPipeline.cs
nvmLantana: Live generation now goes through PopulateManifest too, so CQL IP windows are the org-mapped encounters. Q/NQ still uses every in-period generated encounter. Org_map_scopes_cql_ip_windows_so_unlinked_diagnostic_reports_outside_org_encounters_are_excluded covers it, including the run measure-bundle path.

PR #1873, opened 2026-08-28
LEGLINK-567: Dev: Use Vendor-Secret ID Association for Key Retrieval

smailliwcs:
### 🛠️ Description of Changes

`EpicAuth` derived the Key Vault secret holding a facility's Epic signing PEM by string convention:
`$"{facilityId}-pem"`. That convention was the only thing tying a facility to its key, so a key
could not be shared across facilities and the secret name could not be managed from the Admin UI.

Vendors carry an explicit `SigningKeySecretId` (`VendorAuthenticationSettings.cs:16`), the facility
response exposes its vendor, and `TenantApiService.GetVendorSigningKeySecretId`
(`TenantApiService.cs:84-89`) reads `facility?.Vendor?.Authentication?.SigningKeySecretId`. That
method had **zero production callers** — this change makes it live.

#### `DotNet\DataAcquisition.Domain\Application\Services\Auth\EpicAuth.cs`

- Injected `ITenantApiService` as a sixth constructor parameter (`using
  LantanaGroup.Link.Shared.Application.Services;` added).
- `ResolvePem` now asks the tenant service for the vendor's signing-key secret name and uses it,
  falling back to `$"{facilityId}{PemSuffix}"` only when the lookup returns null/whitespace:

  ```csharp
  var vendorSecretName = await _tenantApiService.GetVendorSigningKeySecretId(facilityId, cancellationToken);

  var pemName = string.IsNullOrWhiteSpace(vendorSecretName)
      ? $"{facilityId}{PemSuffix}"
      : vendorSecretName;
  ```

  The existing "No PEM found in secret manager…" error already interpolates `pemName`, so it names
  whichever secret was actually attempted.
- Threaded the `CancellationToken` that `SetAuthentication` already receives through `GetJwt` →
  `ResolvePem` and used it for the tenant call. The two `ISecretManager.GetSecretAsync` calls keep
  their existing `CancellationToken.None`.
- Failures propagate: the tenant call sits in `GetJwt`, which runs before `SetAuthentication`'s
  `try` block.
- The `PemKeySource.Database` early-return is unchanged, so the tenant service is never called in
  that mode.

**No DI or csproj change was needed.** `ITenantApiService` is already registered
(`GeneralStartupExtensions.cs:320`, `AddTransient`) in the same container that registers `EpicAuth`
(`GeneralStartupExtensions.cs:249`, `AddSingleton`), and both flow into each host via `RegisterAll`.
`DataAcquisition.Domain.csproj` already references `Shared.csproj`.

Lifetime note: a singleton now holds a transient. Safe here — nothing in the `TenantApiService`
graph is scoped (`IHttpClientFactory` and all `IOptions<>` are singletons; `ICreateSystemToken` is
transient over `ILogger` + `IOptions`), so `ValidateScopes` will not trip, and `TenantApiService` is
stateless and creates its `HttpClient` per call.

#### `DotNet\DataAcquisition.AcquisitionWorker\Program.cs`

Added the missing `AddLinkBearerServiceAuthentication` registration, mirroring
`DotNet\DataAcquisition\Program.cs:100-110` so both hosts read the same configuration keys.

Acquisition actually runs in this worker process (`SearchFhirCommand` / `ReadFhirCommand` →
`IAuthenticationRetrievalService` → `EpicAuth`), but the host never called
`AddLinkBearerServiceAuthentication`. `IOptions<LinkBearerServiceOptions>` was therefore
unconfigured and `AllowAnonymous` sat at its `false` default
(`BackendAuthenticationServiceExtension.cs:149`), so the new tenant call would have reached
`CreateSystemToken.ExecuteAsync` with `LinkTokenService:Authority` = `""`
(`AcquisitionWorker/appsettings.json:78`) and thrown `ArgumentNullException`.

The config keys it reads already exist in the worker (`appsettings.json` supplies
`Authentication:Schemas:LinkBearer` and `LinkTokenService`; `appsettings.Docker.json` overrides
`Authentication:EnableAnonymousAccess` to `true`, so Docker takes the anonymous early-return and the
token path is skipped entirely).

No middleware change: the worker only exposes `/health` and `/info`, has no `[Authorize]` endpoints,
and does not call `UseAuthentication`/`UseAuthorization`. Registering the services alone is enough
for `IOptions<LinkBearerServiceOptions>` to be populated for `TenantApiService`. The pre-existing
`AddTransient<ICreateSystemToken, CreateSystemToken>()` at line 35 is now redundant but harmless, so
it was left in place.

#### `DotNet\ServiceTests\UnitTests\DataAcquisition\Services\Auth\EpicAuthTests.cs`

- Added a `Mock<ITenantApiService>` field, passed through the `BuildSut` helper (the only
  construction site), defaulting to `null` via an explicit `.ReturnsAsync((string?)null)` — required
  because a loose Moq returns a null `Task`, not a null `string`. Defaulting to null exercises the
  fallback path, so every pre-existing `-pem` assertion passes unchanged.
- Added a `VendorSecretName = "epic-signing-key"` constant plus an `ISecretManager` stub for it.
- Widened `SetAuthentication_DerivesPemSecretName_FromFacilityId` into
  `SetAuthentication_FallsBackToPemSuffix_WhenVendorSigningKeySecretIdIsNullOrWhitespace`, a theory
  over `null` / `""` / `"   "`.
- New: `SetAuthentication_UsesVendorSigningKeySecretId_WhenTenantServiceProvidesOne` — the vendor
  secret id wins and the `-pem` name is never requested.
- New: `SetAuthentication_PropagatesException_WhenTenantApiServiceThrows` — pins decision 2.
- New: `SetAuthentication_DoesNotQueryTenantService_WhenKeySourceIsDatabase`.

### 🧪 Testing Performed

Tested locally.

### 🧑‍🔬 Unit Testing

- [X] I have written or updated unit tests to cover my changes
- Coverage: 63.1%

### 📓 Documentation Updated

N/A

## Summary by CodeRabbit

- **New Features**
  - Added bearer-service authentication configuration for the acquisition service, including optional anonymous access and token validation settings.
  - Added support for both RSA and ECDSA signing keys when connecting to Epic.
  - Tenant-provided signing-key secret names are now supported, with fallback behavior when unavailable.
  - Local secret retrieval now falls back to environment variables when a stored secret is missing.

- **Bug Fixes**
  - Improved validation and handling of missing authentication audience settings.


DotNet/DataAcquisition.Domain/Application/Services/Auth/EpicAuth.cs
```diff
@@ -166,4 +147,81 @@ private async Task<string> ResolvePem(string facilityId, AuthenticationConfigura
...
+        catch (Exception)
+        {
+            return null;
+        }
+        string? algorithm = ecdsa.KeySize switch
+        {
+            256 => SecurityAlgorithms.EcdsaSha256,
+            384 => SecurityAlgorithms.EcdsaSha384,
+            512 => SecurityAlgorithms.EcdsaSha512,
+            _ => null
```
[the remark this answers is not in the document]
smailliwcs: Resolved in `6621b427f`.

DotNet/DataAcquisition.Domain/Application/Services/Auth/EpicAuth.cs
```diff
@@ -166,4 +147,81 @@ private async Task<string> ResolvePem(string facilityId, AuthenticationConfigura
...
+        {
+            rsa.ImportFromPem(pem);
+        }
+        catch (Exception)
+        {
+            return null;
+        }
+        string? algorithm = rsa.KeySize switch
+        {
+            2048 => SecurityAlgorithms.RsaSha256,
```
arianamihailescu: 1. This changes behavior for existing keys: we've always signed RS256, but now a                                                                                                                                                                                                                                     
  3072/4096-bit key switches to RS384/RS512, and other sizes are rejected.                                                                                                                                                                                                                                 
  Do we know our deployed keys are all 2048-bit — and could this just accept any                                                                                                                                                                                                                                    
  size ≥ 2048 and keep RS256, so nothing changes for existing facilities?                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                    
 2.  Also, the old `Replace("\\r\\n\\t", ...)` workaround is gone — are we sure no                                                                                                                                                                                                                                     
  stored secrets still have literal escape sequences? `ImportFromPem` won't parse those.

DotNet/Shared/Application/Services/SecretManager/LocalSecretManager.cs
```diff
@@ -51,6 +53,10 @@ public LocalSecretManager(
         public Task<string?> GetSecretAsync(string secretName, CancellationToken cancellationToken)
         {
             var secret = _values.GetValueOrDefault(secretName);
+            if (secret == null)
+            {
```
arianamihailescu: Very nice

arianamihailescu (approved):
For RSA support before the alg was always RS256 but now is based on Key size?

DotNet/DataAcquisition.Domain/Application/Services/Auth/EpicAuth.cs
smailliwcs: 1. Resolved in `2595422ee`.  The RSA code path now uses SHA-256 unconditionally for hashing (as before).
 2. Not certain if we have those escape sequences in any stored PEMs, but if we do, IMO we should remove them rather than working around them in code.  (They're no longer valid PEMs if they contain escape sequences.)

